PR #47, opened 2014-08-13
Default python prefix from sys.prefix

vincepii:
With this change, the default installation prefix is taken from the sys module.

This is consistent with how the variable PYTHON_SITELIB is defined.

Without this change, on a system using /usr/local as install prefix, the setup.py script will use /usr and the chmod command will use /usr/local, resulting in an error.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ endif
 MANDIR=/usr/share/man
 
 ifndef PREFIX
-  PREFIX=/usr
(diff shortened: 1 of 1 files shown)


tomjelinek:
Thank you for your patch. It has been cherry-picked https://github.com/feist/pcs/commit/dde1199887b235d50a4d459f0056c91209768d8d

PR #48, opened 2014-09-12
Fix pcsd RHEL6/C6 compatibility

deriamis:
This commit would fix both pcs and pcsd compatibility with RHEL 6/CentOS 6 when the required software versions are actually installed (such as with a custom RPM):

1) All checks for "release 6" via /etc/redhat-release were converted to checks for required software versions instead.
2) The logic for "is_rhel6" was reversed to "is_rhel7_compat"
3) All references to systemd or systemctl were placed behind conditionals specific to systemd being installed. Some of these were previously if_rhel6 checks.
4) pcsd/ssl.rb has been modified so daemonizing pcsd will now work on both RHEL 6 and RHEL 7. On RHEL 6, the systemd callback is not used and WEBRick::Daemon::start is called instead so it forks properly.
5) The top-level Makefile has been altered to check for the presence of systemctl when installing init files so it won't install systemd stuff if it's not there, but it will install the init script in that case.
6) pcsd/Makefile was modified so it considers the Ruby gems a dependency of build_gems (which won't try downloading gems), and it doesn't try to differentiate between RHEL 6 and RHEL 7 gems anymore.
7) pcsd/Gemfile (and pcsd/Gemfile.lock) have been modified so that the one problematic gem in RHEL 6 (rpam-ruby19-feist) is now pulled from a git repository instead.
8) Some dependencies were not being included in pcsd/Gemfile (and pcsd/Gemfile.lock), and they have been added so the build from makefile works correctly.
8) pcsd/pscd.rb was modified so it requires bundler/setup and executes Bundler.setup(:default), which allows rpam-ruby19-feist to be used (since rubygems won't do it).
9) pcsd/pcsd (the init script) was modified to execute via bundler rather than directly with ruby, which allows the bundled gems to be required properly (needed for rpam-ruby19-feist).
10) A few text changes to errors so the compatibility check is more obvious (saying "this isn't RHEL 7" versus "you're using X software").

I know this is a large commit, but I really did not see any way of breaking it up into something more manageable in this case. I have tested this on a RHEL 6 box, but I don't have access to a RHEL 7 box at the moment to do further testing. I don't imagine any of the changes I made would be a problem for that, though. Please do let me know if you have any questions or suggestions!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,10 +8,23 @@ ifeq ($(PYTHON_SITELIB), /usr/lib/python2.7/dist-packages)
   EXTRA_SETUP_OPTS="--install-layout=deb"
 endif
 
+IS_SYSTEMCTL=false
+ifeq ("$(wildcard /usr/bin/systemctl)","/usr/bin/systemctl")
+  IS_SYSTEMCTL=true
+else
+  ifeq ("$(wildcard /bin/systemctl)","/usr/bin/systemctl")
+    IS_SYSTEMCTL=true
+  endif
+endif
+
 MANDIR=/usr/share/man
 
 ifndef PREFIX
   PREFIX=/usr
+endif
+
+ifndef initdir
+  initdir=/etc/init.d
 endif
 
 install: bash_completion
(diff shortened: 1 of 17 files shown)


tomjelinek:
Thank you very much for this patch and sorry for the delay.
1) We need to take a closer look at this. Running corosync and pacemaker to determine their version is much slower than reading a file.
2) I'm not going to accept these changes. We focus mainly on latest software releases so we want any code dealing with older software to be a special case and not the other way around. Therefore it's better to have is_rhel6 function and possibly add is_rhel7 function when working on RHEL 8.
3) Great, this has been cherry-picked https://github.com/feist/pcs/commit/f1a8b4825775ba1a5ab12e843e7523348e40d94b
4) Partially cherry-picked https://github.com/feist/pcs/commit/f1a8b4825775ba1a5ab12e843e7523348e40d94b
5) Great, this has been cherry-picked https://github.com/feist/pcs/commit/f1a8b4825775ba1a5ab12e843e7523348e40d94b
6) to 9) We will deal with these changes later.
10) cherry-picked https://github.com/feist/pcs/commit/639ed58737a42bde8e2be98759f3ab5200142185

tomjelinek:
I did a review of the remaining proposed changes and implemented some of them in https://github.com/feist/pcs/commit/ecfdc731dc846b10bdb475e3b4f44900cfc9badc. As bundler is not available on RHEL 6 as a RPM package I'm not going to accept some changes related to it.


> 6) pcsd/Makefile was modified so it considers the Ruby gems a dependency of build_gems (which won't try downloading gems), and it doesn't try to differentiate between RHEL 6 and RHEL 7 gems anymore.

Accepted. Original RHEL 6 line left in the Makefile for anyone who doesn't have bundler installed.


> 7) pcsd/Gemfile (and pcsd/Gemfile.lock) have been modified so that the one problematic gem in RHEL 6 (rpam-ruby19-feist) is now pulled from a git repository instead.

Accepted. I changed the location of the gem though.


> 8) Some dependencies were not being included in pcsd/Gemfile (and pcsd/Gemfile.lock), and they have been added so the build from makefile works correctly.

Added missing json dependency. Sinatra-reloader is part of sinatra-contrib gem so I'm not adding it.


> 8) pcsd/pscd.rb was modified so it requires bundler/setup and executes Bundler.setup(:default), which allows rpam-ruby19-feist to be used (since rubygems won't do it).

Rejected, bundler is not available on RHEL 6 as a RPM package and it works for me just fine without accepting the change.


> 9) pcsd/pcsd (the init script) was modified to execute via bundler rather than directly with ruby, which allows the bundled gems to be required properly (needed for rpam-ruby19-feist).

We are going to make more changes to the init script later as there are issues with it anyways.

PR #50, opened 2014-12-11
create man page on install /tarball targets

grooverdan:


tomjelinek:
Thank you for your patch. It would be nice to have the man page generated automatically from the help. Unfortunatelly help2man produces corrupted man page from the current pcs help. That's why I'm not going to accept your patch.

PR #55, opened 2015-03-13
Rikkotec Github.com and pcsd Makefile Patch For GNU/Linux

rikkotec:
**Source Updates**
- Provided support for building pcsd on GNU/Linux, specifically Debian.  See `pcsd/Makefile`
- Added `README.md` to the source, which Github favors over README - and which presents a prettier and more readable documentation via the Github.com web project interface with updated documentation for GNU/Linux users in the pcsd installation steps.
- Updated `README` file to include additional information for GNU/Linux users in the pcsd installation steps.

**Additional Notes**
- Tested on Debian Jessie, with Pacemaker 1.11 and Corosync 2.x from experimental.
- I did not use sudo for `make get_gems` in the pcsd directory, this worked fine as non-root.
- I had to run `sudo make install`, and `sudo make install_pcsd` in order  to get privileges to create directories and copy files.
- I had to run `sudo systemctl daemon-reload` before I could utilize pcsd via `systemctl start pcsd.service`
- A warning is displayed when you `sudo make install_pcsd` from ruby, explaining not to install gems as the
  root user...this was unavoidable for me - but I'm new to Ruby, perhaps there is a way to avoid that.
- Using `sudo` should not be too much of a problem on Debian, as you can only use Pacemaker and
  Corosync as root anyhow.  File permission changes would be necessary through the whole cluster stack.
- I left an additional `else` in the `ifeq` blocks of `pcsd/Makefile` for Debian, in order to provide others a sort of *place-holder* for helping them to support additional Debian versions, since I didn't do any extensive testing.

### DIFF
--- a/README
+++ b/README
@@ -33,4 +33,7 @@ You can also install pcsd which operates as a GUI and remote server for pcs.  To
 # cd pcsd ; make get_gems ; cd ..
 # make install_pcsd
 
+If you are using GNU/Linux its now time to:
+# systemctl daemon-reload
+
 If you have an questions or concerns please feel free to email cfeist@redhat.com or open a github issue on the pcs project.
--- a/README.md
+++ b/README.md
@@ -1,0 +1,86 @@
+## PCS - Pacemaker/Corosync configuration system
+
+
+### Quick Start
+***
+
+
+- **PCS Installation from Source**
+
+   Run the following in terminal:
+  
+   ```shell
+   # tar -xzvf pcs-0.9.138.tar.gz
+   # cd pcs-0.9.138
+   # make install
+   ```
+
+   This will install pcs into `/usr/sbin/pcs`.
+
+<br />
+- **Create and Start a Basic Cluster**
+
+   To create a cluster run the following commands on all nodes (replacing node1, node2, node3 with a list of nodes in the cluster).
+
+   ```shell
+   # pcs cluster setup --local --name cluster_name node1 node2 node3
+   ```
+
+   Then run the following command on all nodes:
+
+   ```shell
+   # pcs cluster start
+   ```
+
+<br />
+- **Check the Cluster Status**
+
+   After a few moments the cluster should startup and you can get the status of the cluster
+
+   ```shell
+   # pcs status
+   ```
+
+<br />
+- **Add a Cluster Reosurce**
+
+   After this you can add resources and stonith agents:
+
+   ```shell
+   # pcs resource help
+   ```
+
+   and
+
+   ```shell
+   # pcs stonith help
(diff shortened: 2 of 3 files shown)


rikkotec:
Oh well there is one change you'll probably want to make:

    pcsd may also necessary in order to follow the guides on the clusterlabs.org website.

It's in the new `README.md` file, and it should read:

    pcsd may also be necessary in order to follow the guides on the clusterlabs.org website.

And I did not add that line to `REAMDE` either, if you wanted to do that as well.

</br>
I knew I missed something ><

rikkotec:
So I'm going to leave this here so that you may cherry-pick from it if you like. Though I've decided to package for Debian this project -> as I've had to build the entire cluster stack from source due to the shape and outdated state of Debian's packages. I guess this is one of those *itches*

So with that said, if you just want to take the README.md and reject the pull request, I can just add the necessary rules to debian/rules in the package I'll create, and provide a proper init there as well.

Though, if you want to take that too and make it work to build from source on Debian as well -> I'll make another commit on this pull-request with the changes stated above, and more;  within the next couple of days - tested.

Best,

tomjelinek:
Thank you for the fix and notes.

I cherry-picked your changes in README files.
Proposed Makefile fix seems to be more complex than necessary so I'm not going to use that. I made a simple fix instead. Checking for other distributions may be added when needed.

Changes of /etc/init.d/pcsd seem to be incompatible with RHEL. We do not want to break compatibility with one distribution when adding support for another.

rikkotec:
@tomjelinek 

> Thank you for the fix and notes.

No problem, it was more notes than fixing - but still :)

> Proposed Makefile fix seems to be more complex than necessary so I'm not going to use that. I 
> made a simple fix instead. Checking for other distributions may be added when needed.

The makefile I wasn't sure about -> I didn't think it needed to be that involved but I don't know enough about RHEL to have made a better fix without knowing Debian support was there and not breaking RHEL.

> Changes of /etc/init.d/pcsd seem to be incompatible with RHEL. We do not want to break compatibility 
> with one distribution when adding support for another.

The init script I know is dead wrong - even for Debian (my edit that is). I saw in the other pull-request that you cherry-picked; so I decided to submit pull request assuming you might review notes/cherry-pick things you wanted since it wasn't so huge of a difference.

I'm actually packaging pcs for Debian as we speak. I can introduce makefile rules in debian/rules (but if you fixed it then probably wont need to do that). And a new init script can be used specific to Debian via the package as well, without messing with upstream source.

Thanks for your reply, I'll give it a test if the changes to the makefile are updated in master :)

PR #59, opened 2015-05-07
Streamline upstream for compatibility with Debian

rikkotec:
- Updated the root Makefile to include checks for Debian and 
   modify the installation appropriately.
 - The set of changes should allow for a proper 'from source' 
   installation when downloaded and lacks any Debian 
   packaging modifications, meant for a pull request to be 
   sent upstream
   
Summary of Changes for Debian:
 - Arch indep files in /usr/share
 - Sysconfig is /etc/default
 - system-auth is /etc/pam.d/common-*; Debian specific pam.d 
   configuration, pcsd.pam.debian, is added
 - Debian specific systemd file, pcsd.service.debian, is added
 - Debian specific init; pcsd.debian is added
 - Variables are introduced which discretely eases packaging;
   BUILD_GEMS, systemddir
 - pcs/settings.py has variants for multi-arch support with 
   64-bit and 32-bit architectures, updated to know the 
   proper location of pacemaker binaries depending on the 
   flag install_settings during build

Change-Id: Ie06f5ed5232fbf00881f7c4a45c00a7822645d4c
Signed-off-by: Richard B Winters <rik@mmogp.com>


rikkotec:
Sorry, but my OCD just won't let me leave the redunant else in the Makefile, so closing yet again to redo the pull request

PR #62, opened 2015-05-08
Streamline upstream for compatibility with Debian

rikkotec:
Sent it again, this time with changes based on team reviewed feedback, especially regarding the check for systemctl.

Updated the root Makefile to include checks for Debian and modify the installation appropriately.

The set of changes should allow for a proper 'from source' installation when downloaded and lacks any Debian packaging modifications, meant for a pull request to be sent upstream
   
Summary of Changes for Debian:
 - Arch indep files in /usr/share
 - Sysconfig is /etc/default
 - system-auth is common-*; Debian specific pam.d configuration, pcsd.pam.debian, is added
 - Debian specific systemd file, pcsd.service.debian, is added
 - Debian specific init; pcsd.debian is added
 - Variables are introduced which discretely eases packaging;  BUILD_GEMS, install_settings, systemddir, and SYSTEMCTL_OVERRIDE
 - Check for systemd fixed for debian, additionial variable for discretely easing packaging  was added
 - pcs/settings.py has variants for multi-arch support with 64-bit and 32-bit architectures, updated to    proper location of pacemaker binaries depending on the flag install_settings during build


Makefile
```diff
@@ -1,3 +1,21 @@
+# Compatibility with GNU/Linux [i.e. Debian] based distros
+UNAME_OS_GNU := $(shell if uname -o | grep -q "GNU/Linux" ; then echo true; else echo false; fi)
+UNAME_KERNEL_DEBIAN := $(shell if uname -v | grep -q "Debian\|Ubuntu" ; then echo true; else echo false; fi)
```
rikkotec: Added `\|` to line 2, so that grep checks for Debian OR Ubuntu

pcsd/pcsd.debian
```diff
@@ -0,0 +1,141 @@
...
+
+run()
+{
+  if is_running; then
+    PID="$(cat $PIDFILE)"
+    echo "Daemon is already running as PID $PID"
+    return 1
+  fi
+
+  nohup $DAEMON $DAEMON_ARGS > /dev/null 2>&1 &
```
rikkotec: I added `&` to line 73 as requested by @roccolord, it did test ok for me.

tomjelinek:
Thank you very much for the patch.

I did more fixes for Debian based on your changes https://github.com/feist/pcs/commit/b31f639b6581dab7162355521aa3e69f52d89654. It would be great if you could take a look at it or test it:
* Added pcsd location to settings.py and Debian specific variants
* Added settings.rb file and Debian specific variants
* Updated the root Makefile to install Debian specific version of settings.rb
* Updated check for systemd in bootstrap.rb and utils.py based on paths from the root Makefile

Issue #71, opened 2015-12-16
PCS only looks in the installed packages path (/usr/lib/systemd/system) for unit files, ignoring local-only units in /etc/systemd/system


loz-hurst:
From the systemd man pages (man systemd.unit):

Unit File Load Path

Unit files are loaded from a set of paths determined during compilation, described in the two tables below. Unit files found in directories listed earlier override files with the same name in directories lower in the list.

When the variable $SYSTEMD_UNIT_PATH is set, the contents of this variable overrides the unit load path. If $SYSTEMD_UNIT_PATH ends with an empty component (":"), the usual unit load path will be appended to the contents of the variable.

Table 1.  Load path when running in system mode (--system).

Path	Description
/etc/systemd/system	Local configuration
/run/systemd/system	Runtime units
/usr/lib/systemd/system	Units of installed packages

loz-hurst:
Local units (i.e. services that are on the system but not as a result of package installs), e.g. to control binary-only servers in /opt, should be in /etc/systemd/system and not /usr/lib/systemd/system but pcs only checks in /usr/lib/systemd/system in utils.is_valid_resource despite pcs resource list showing all the units, including those only in /etc/systemd/system.

utils.is_valid_resource should also check for units in /etc/systemd/system.

tomjelinek:
Fixed by patch from https://github.com/feist/pcs/pull/72 in https://github.com/feist/pcs/commit/e56ebe1ceec7e6557dfb35d9e3b651078405971f. Thanks for reporting this and posting the patch.

kubikrobert:
One location is still missing: /lib/systemd/system (Ubuntu 16.04)

tomjelinek:
@kubikrobert See https://github.com/ClusterLabs/pcs/issues/164

PR #72, opened 2015-12-16
Fix for local systemd resources (issue #71)

loz-hurst:
Added check /etc/systemd/system for local services in is_valid_resource

Stops  is_valid_resource returning False for local resources.


loz-hurst:
Sorry, issue #71 not #17 - my bad!

tomjelinek:
Thanks for the patch! I accepted it to upstream in https://github.com/feist/pcs/commit/e56ebe1ceec7e6557dfb35d9e3b651078405971f

PR #97, opened 2016-05-20
Support for using login functions in AJAX

marxsk:
This commit fixes two following problems:

* Successful /logout redirects you to /login what is not needed for AJAX calls.
* Endpoint where you can check that your session cookie is valid did not exist, now
  it is possible to use /login-status
* Random number is returned instead of cookies ID
* /login-status works also outside of AJAX

### DIFF
--- a/pcsd/pcsd.rb
+++ b/pcsd/pcsd.rb
@@ -83,7 +83,7 @@ before do
     $session_storage_env = env
   end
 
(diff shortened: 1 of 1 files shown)


tomjelinek:
This has been cherry-picked in https://github.com/feist/pcs/commit/06c3ed7936c0c3b09f7f6c33c1b4023510660bdd

PR #101, opened 2016-07-28
Coloring pcs status

dchirikov:
Hi,

here is a patch which adds fancy colors and effects to `pcs status` output.
It is expected, that it will enhance usability.

New form of output will be available by `pcs status --color`.


tomjelinek:
Hi,

Thank you for the patch but I am not going to merge it upstream.

I agree that coloring pcs output enhances usability. This kind of coloring should be done in crm_mon though as it cannot reliably work in pcs - there may be false positives for example.

PR #102, opened 2016-08-12
fix disable_service

antblant:
When try to add new node to cluster with Ubuntu Server 16.04.1:

```
# pcs cluster node add ubuntu-4 --enable --start
Disabling SBD service...
Error: debian-4: Disabling SBD failed
```

because the current version of systemctl fails when you try to disable nonexistent service:

```
ubuntu-4 # systemctl disable sbl.service
Failed to execute operation: No such file or directory
ubuntu-4 # echo $?
1
```

We need to check if service exist also for systemctl.

### DIFF
--- a/pcsd/pcs.rb
+++ b/pcsd/pcs.rb
@@ -1995,14 +1995,14 @@ def enable_service(service)
 end
 
 def disable_service(service)
-  if ISSYSTEMCTL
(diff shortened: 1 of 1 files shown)


tomjelinek:
Thank you for the patch!

PR #103, opened 2016-08-23
Export variables from /etc/default/pcsd

vvidic:
Similar change exists in RHEL init script for loading /etc/sysconfig/pcsd


tomjelinek:
Thanks for the patch!

PR #104, opened 2016-08-23
Self restart fails when using Debian init script

vvidic:
The child process created in pcsd_restart() still listens on the TCP port so the restart fails with:
   /usr/lib/ruby/2.3.0/socket.rb:205:in `bind': Address already in use - bind(2) for [::]:2224 (Errno::EADDRINUSE)

The fix is to call the restart commands via exec as this will close the listening socket automatically.


tomjelinek:
Thanks for the patch!

PR #121, opened 2017-01-05
Tests: handle change in an error message in lxml 3.7

AdamWill:
I guess we could also do this as a constant rather than a function...if you want me to change that, just yell. This is needed to make pcs build properly (with test suite passing) on Fedora Rawhide (and F24 and F25, now, as they also have lxml 3.7).

### DIFF
--- a/pcs/lib/commands/test/test_resource_agent.py
+++ b/pcs/lib/commands/test/test_resource_agent.py
@@ -8,7 +8,7 @@ from __future__ import (
 import logging
 from lxml import etree
 
-from pcs.test.tools.assertions import assert_raise_library_error
+from pcs.test.tools.assertions import assert_raise_library_error, start_tag_error_text
 from pcs.test.tools.custom_mock import MockLibraryReportProcessor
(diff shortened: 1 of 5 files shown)


AdamWill:
Gah! The difference here doesn't *seem* to be lxml 3.7, because this works on Fedora Rawhide and 25, but fails on 24; 24 has lxml 3.7.0. 25 and Rawhide have 3.7.1, but upstream changelog says nothing in the source was changed between 3.7.0 and 3.7.1 (the release was just to fix some binary package issues).

So we need to figure out what the difference really is...the "Start tag expected, '<' not found" error ultimately comes from libxml, but I guess I get to dig into exactly how it gets down to this `LibraryError` class to figure out why the message is changing...

tomjelinek:
Thank you very much for the patch!

I think your current solution is just fine. We're not really interested in the error message, pcs just passes it to the user. Therefore tests are the right place to fix it.

Cherry-picked in https://github.com/ClusterLabs/pcs/commit/f98b07d5025904e451e265e3ac93211fdba289da

PR #122, opened 2017-01-05
Fix tests for Python 3.6

AdamWill:
We need to set a proper value for `env.booth.name` in two more
tests, or Python chokes trying to `os.path.join()` a MagicMock.

I'm assuming this is a Python 3.6 thing, but didn't actually check; it's needed to make pcs build correctly (with tests passing) on current Fedora Rawhide. Without this change, the tests fail in `get_config_file_name`, because `name` is a MagicMock.


tomjelinek:
Thank you very much for the patch!

PR #130, opened 2017-03-22
cli: skip superfluous text -> bytes conversion on clufter's output

jnpkrn:
Respective outputs are already represented with bytes (as 'bytestring'
protocol suggests), and moreover, such bytes cannot be "encoded" in
Python 3 otherwise leading to AttributeError (bytes object has no
attribute encode) exception.

Note that this is only relevant in Python 3 context, which pcs uses
by default in python3-charged distros (e.g. Fedora) and which is now
supported in clufter (which enabled a discovery of this discrepancy):
http://oss.clusterlabs.org/pipermail/users/2017-March/005323.html

### DIFF
--- a/pcs/config.py
+++ b/pcs/config.py
@@ -698,14 +698,14 @@ def config_import_cman(argv):
         config_backup_add_version_to_tarball(tarball)
         utils.tar_add_file_data(
             tarball,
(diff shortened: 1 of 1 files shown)


jnpkrn:
Tested with: `pcs config {import-cman,export pcs-commands*}` in Fedora rawhide

PR #143, opened 2017-08-25
Check if systemd is currently running (Fixes #118)

### LINKED ISSUE #118: pcs trying to use systemctl even though sysvinit is the init system
System: Debian 8 Jessie

In Debian's case it is perfectly fine to have multiple init systemd installed at the same time.
The is_systemctl check performed in lib/external.py assumes that if there are systemd's binaries installed in the system, then it must be running systemd as init system.

Steps to reproduce:

1. Install sysvinit-core 
2. Reboot the system
3. Try starting the cluster
4. Receive the following error when running pcs cluster start:

```
# pcs cluster start
Starting Cluster...
Failed to get D-Bus connection: Unknown error -1
Error: unable to start corosync
# /etc/init.d/corosync status
corosync is running.
```

To workaround this issue, one could simply remove systemd from the system, but it's inconvenient.
```
# apt remove systemd -y
# pcs cluster start
Starting Cluster...
```

Proposed solutution:
Check if systemd is used as the init system. Sysvinit will always have pid = 1, whereas in systemd's case it appears to be somewhat random:

```
$ pidof systemd
6906
```

```
# pidof systemd
# pidof init
1
```

vvidic:
Rather than just checking if systemctl binary is installed,
verify that systemd is currently running as init.

### DIFF
--- a/pcs/lib/external.py
+++ b/pcs/lib/external.py
@@ -94,14 +94,12 @@ def is_systemctl():
     Check whenever is local system running on systemd.
     Returns True if current system is systemctl compatible, False otherwise.
     """
-    systemctl_paths = [
-        '/usr/bin/systemctl',
-        '/bin/systemctl',
-        '/var/run/systemd/system',
+    systemd_paths = [
         '/run/systemd/system',
+        '/var/run/systemd/system',
     ]
-    for path in systemctl_paths:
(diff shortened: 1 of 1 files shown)


tomjelinek:
Thanks for the patch.

Issue #145, opened 2017-09-29
Idempotency issue when updating a guest node


actatux:
There is an idempotency issue with the pcs resource update command when using meta parameter remote-node.

 Reproducer with the new EL 7.4 commands:
```
[root@rhel74 ~]# pcs resource create myguest ocf:heartbeat:VirtualDomain \
 config=/var/lib/libvirt/images/myguest/myguest.xml \
 force_stop=false hypervisor=qemu:///system \
 op monitor interval=60 on-fail=restart timeout=90 \
 op start interval=0s timeout=120 \
 op stop interval=0s timeout=90 \
 meta failure-timeout=60 migration-threshold=2

[root@rhel74 ~]# pcs cluster node add-guest myguest.domain myguest --skip-offline

[root@rhel74 ~]# pcs resource show myguest
 Resource: myguest (class=ocf provider=heartbeat type=VirtualDomain)
  Attributes: config=/var/lib/libvirt/images/myguest/myguest.xml force_stop=false hypervisor=qemu:///system
  Meta Attrs: failure-timeout=60 migration-threshold=2 remote-node=myguest.domain
  Operations: monitor interval=60 on-fail=restart timeout=90 (myguest-monitor-interval-60)
              start interval=0s timeout=120 (myguest-start-interval-0s)
              stop interval=0s timeout=90 (myguest-stop-interval-0s)

[root@rhel74 ~]# pcs resource update myguest meta remote-node=myguest.domain
Warning: this command is not sufficient for creating a guest node, use 'pcs cluster node add-guest'
Error: 'myguest.domain' already exists
```
The resource update call graph is:

resource_update() -> _detect_guest_change() -> _validate_guest_change() -> guest_node.validate_conflicts() -> does_id_exist(tree, node_name)

> def does_id_exist(tree, check_id):
>     """
>     Checks to see if id exists in the xml dom passed
>     tree cib etree node
>     check_id id to check
>     """

I am trying to update the resource with the same remote-node value, so the id exists in the xml dom and the update command fails.

 This check makes sense to ensure we cannot define the same remote-node for several resources.

However, the command should be idempotent when we update a resource with the same remote-node value. Our configuration tools generate the update commands from external configuration files. We build the pcs resource update command with all the parameters for the resource to ensure that any modification made outside of the configuration tool is reverted to the external configuration file value.

idevat:
Thank you for reporting and fixing this issue.

PR #146, opened 2017-09-29
Make resource update idempotent with remote-node

actatux:
If the remote-node name argument is already defined for the resource,
do not deal with guest change.

Fix issue #145.


idevat:
Thank you for reporting and fixing this issue.

PR #148, opened 2017-10-13
Include PCSD_DEFAULT_PORT in pcsd/settings.rb.debian

vvidic:
pcsd service fails to start without the default value.


tomjelinek:
Thanks, merged in https://github.com/ClusterLabs/pcs/commit/426009739383227797e6426e456afff60e20efd6.

PR #149, opened 2017-10-13
Enable python3 build on Debian

vvidic:
Use correct install directory for libs.


tomjelinek:
Thanks, merged in https://github.com/ClusterLabs/pcs/commit/15e3815ad8047af025c276620b7010b638b43afc.

PR #155, opened 2017-11-27
pcsd.service files: fix bogus ExecStart specification

jnpkrn:
Command Lines section of systemd.service(5) makes it clear that

  redirection [...], running programs in the background using "&"
  [...] elements of shell syntax are not supported

for Exec family of unit file directives.

Backgrouding is simply a non-sense while ">/dev/null" may be achieved
with StandardOutput=null[*], otherwise such output is forwarded to
journal, or whatever is configured as a system-wide default.
In my testing, nothing is produced on standard output (nor standard
error output, stderr, for that matter) till it gets internally (along
with stderr) redirected to pcsd native log file (/var/log/pcsd/pcsd.log)
later in the startup sequence.  Hence do nothing special about it.

Fortunately, the starting script did not react on anything passed as
arguments (e.g. as happened to "&" token), leaving this sanitization
merely for good measure.

[*] /dev/null redirection would get overridden later on as mentioned,
    regardless

### DIFF
--- a/pcs/snmp/pcs_snmp_agent.service
+++ b/pcs/snmp/pcs_snmp_agent.service
@@ -4,7 +4,7 @@ Requires=snmpd.service
 
 [Service]
 EnvironmentFile=/etc/sysconfig/pcs_snmp_agent
-ExecStart=/usr/lib/pcs/pcs_snmp_agent > /dev/null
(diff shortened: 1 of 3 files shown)


tomjelinek:
Thanks for the patch!

PR #161, opened 2018-02-15
update information about default token timeout in CMAN clusters.

OndrejHome:
This is just a documentation fix.

In CMAN clusters it is 10s while elsewhere it is 1s.
This is relevant for CentOS/RHEL 6.X.
When cluster is created in CentOS/RHEL 6.X without specifying token timeout then 10 seconds is used and can be verified with command below.

~~~
# corosync-objctl |grep totem.token
cluster.totem.token=10000
totem.token=10000
~~~


tomjelinek:
This has been cherry-picked in https://github.com/ClusterLabs/pcs/commit/56313b1e656b04427be1d534d740e4c23b0f493e.
Thanks for the patch.

PR #162, opened 2018-03-08
Check if node name stored in corosync.conf

AlekseyChudov:
According to https://clusterlabs.org/pacemaker/doc/en-US/Pacemaker/1.1/html/Pacemaker_Explained/s-node-name.html Pacemaker Gets the Node Name value stored in corosync.conf under name in the nodelist if ring0_addr contain an IP address.

E.g.
```
nodelist {
    node {
        ring0_addr: 172.31.200.1
        nodeid: 1
        name: srv1
    }

    node {
        ring0_addr: 172.31.200.2
        nodeid: 2
        name: srv2
    }
}
```

It is very useful in multihomed network configuration or when you do not want to rely on /etc/hosts and/or DNS.

This patch eliminates "WARNING: corosync and pacemaker node names do not match (IPs used in setup?)" in this case.

### DIFF
--- a/pcs/utils.py
+++ b/pcs/utils.py
@@ -1994,11 +1994,18 @@ def getCorosyncNodesID(allow_failure=False):
             node_list_node_mapping[m.group(1)] = m.group(2)
 
     for line in output.rstrip().split("\n"):
-        m = re.match("nodelist\.node\.(\d+)\.ring0_addr.*= (.*)", line)
-        # check if node id is in node_list_node_mapping - do not crash when
-        # node ids are not specified
+        # check if node name stored in corosync.conf under name in the nodelist
+        m = re.match("nodelist\.node\.(\d+)\.name.*= (.*)", line)
         if m and m.group(1) in node_list_node_mapping:
             cs_nodes[node_list_node_mapping[m.group(1)]] = m.group(2)
+            continue
+
+        # otherwise use ring0_addr as node name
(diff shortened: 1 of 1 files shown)


tomjelinek:
Thank you for the patch. I'm not going to accept it, however. The reason is that it removes a warning while it doesn't deal with reasons the warning is there in the first place.

You see, there is quite a lot of commands in pcs which takes nodes as parameters. The problem is, some of the commands work with pacemaker (e.g. pcs node maintenance, pcs node standby, pcs constraint location), others with corosync (e.g. pcs cluster setup) and others with nodes itself (e.g. pcs cluster auth, pcs cluster start). Pacemaker commands need the nodes to be specified the way pacemaker knows them (i.e. by name) while corosync commands need the nodes to be specified as seen by corosync (i.e. by ring0_addr). And pcs itself uses ring0_addr as the node identifier. This is quite confusing and it is the reason the warning has been put there, see https://bugzilla.redhat.com/show_bug.cgi?id=1126998. Long story short, pcs currently does not support node names in corosync.conf and the preferred way is to set ring0_addrs so they match node names in pacemaker.

I totally agree with you that specifying node names in corosync.conf is a very useful feature. We are currently working on full node names support in pcs. It is, however, a big change both from users and code point of view, spanning quite a few pcs commands and changes one of the basic principles of pcs - nodes will be identified by names instead by ring0_addr. That is why, among other reasons, it is targeted for pcs-0.10 branch.

AlekseyChudov:
I see. Thanks for the exhaustive explanation.

PR #166, opened 2018-04-12
fix 'cib-push' exit code when the diff is empty

actatux:
pcs cib-should not consider an empty diff as an error. Instead, we notice
the user and exit with 0.

### DIFF
--- a/pcs/cluster.py
+++ b/pcs/cluster.py
@@ -1466,9 +1466,10 @@ def cluster_push(argv):
         if error.strip():
             utils.err("unable to diff the CIBs:\n" + error)
         if not patch.strip():
-            utils.err(
+            print(
                 "The new CIB is the same as the original CIB, nothing to push."
             )
+            sys.exit(0)
 
         command = ["cibadmin", "--patch", "--xml-pipe"]
         output, error, retval = runner.run(command, patch)


tomjelinek:
Summary of the preceding IRC discussion:
The use case here seems to be that an automated process (puppet module) consider it has to apply some diff while there are no diff. In such a case, pcs should exit with 0 so that the puppet module considers its task to be completed successfully.
The argument for pcs to exit with 1 in the case of an empty diff is that its task is to change something in the CIB and when there is no change, it should be reported. An empty diff may mean the command was used incorrectly or commands run before it didn't do any changes. It's up to the pcs user to decide if the situation is OK or not.


I've been thinking about this and got to a conclusion that pcs should exit with 0 in the case of an empty diff, the current status is in fact a bug. Reasoning: `pcs node unstandby` exits with 0 no matter if the node was in standby mode or not (considering no errors occur). It is pushing the CIB as a diff internally. When the node is not in standby mode, the diff is empty, yet pcs doesn't exit with 1. When running the same command on a CIB in a file and then pushing it as a diff manually, it should work the same and exit with 0. The same applies for other pcs commands.


Cherry-picked to both pcs-0.9 https://github.com/ClusterLabs/pcs/commit/be7fa7b28fd08571723164d90f07e414862cab99 and pcs-0.10 https://github.com/ClusterLabs/pcs/commit/4eb8c8f5163b0da238054ef6c307c7194a33a643. Thanks for bringing this up!

PR #171, opened 2018-05-03
Man: pcs.8: primarily refer to the new pacemaker-2.0 daemon man pages

jnpkrn:
Still, allow one to fall back to the original names conveniently,
which remains important for several years to come.

Also normalize the relevant punctuation.

### DIFF
--- a/pcs/pcs.8
+++ b/pcs/pcs.8
@@ -460,13 +460,16 @@ Remove the permission id specified (permission id's are listed in parenthesis af
 .SS "property"
 .TP
 [list|show [<property> | \fB\-\-all\fR | \fB\-\-defaults\fR]] | [\fB\-\-all\fR | \fB\-\-defaults\fR]
-List property settings (default: lists configured properties).  If \fB\-\-defaults\fR is specified will show all property defaults, if \fB\-\-all\fR is specified, current configured properties will be shown with unset properties and their defaults.  Run 'man pengine' and 'man crmd' to get a description of the properties.
+List property settings (default: lists configured properties).  If \fB\-\-defaults\fR is specified will show all property defaults, if \fB\-\-all\fR is specified, current configured properties will be shown with unset properties and their defaults.
+See \fBpacemaker-controld\fR(7) and \fBpacemaker-schedulerd\fR(7)  man pages for a description of the properties.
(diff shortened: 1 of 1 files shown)


kgaillot:
Good catch.

I'm thinking pcs should reference the new names only. IIRC pcs has somewhat strict dependencies on particular versions of pacemaker, though I'm not sure if the version compatibility is documented anywhere. But since pcs will likely change to create the new promotable clone syntax understandable only by Pacemaker 2.0+, it should be able to assume the new names will be in use here.

tomjelinek:
There are two pcs branches, pcs-0.9 and pcs-0.10 which support pacemaker 1.x only and pacemaker 2.x only respectively. Pcs-0.10 will become the master branch once it is ready. There are currently no plans to have a single pcs version supporting both pacemaker 1.x and 2.x. This is documented in pcs readme and we will make that information more visible when moving pcs-0.10 to master.

jnpkrn:
Refreshed so that original daemon names are not stated.

jnpkrn:
Should I interpret closing this PR as just non-transparent on-the-right-track
part of the process or a rejection (without a word?).

tomjelinek:
It was not closed intentionally. It happened automatically when the pcs-0.10 branch was made the master branch and was itself deleted. I did not realize this would happen. Thanks for pointing this out. Sadly, github does not allow to reopen the pull request. Could you update the PR?

jnpkrn:
True, seems to be the case at least since early 2016:
https://stackoverflow.com/a/32658536

Another reminder and lessons learned that branching model should be
given careful considerations from the get-go and not be messed up with
later on, because of all these unintended side-effects.  Not sure, but
merging `pcs-0.10` to master explicitly first might have been the best
execution of "pcs-0.10 will become the master branch" (unless it was
plain fast-forward case, which I think it wasn't, because why to
branch off initially, then?).

Also not sure how, e.g., GitLab behaves in such circumstances, but
if more gracefully, it would be yet another incentive to move away
from GitHub as a primary project portal [1].

Enough of rumbling about the inflicted inconvenience,
new PR for this is #176.

[1] https://lists.clusterlabs.org/pipermail/developers/2018-June/001241.html

PR #174, opened 2018-06-24
fix syntax for python3

vvidic:


idevat:
Merged. Thanks for the patch.

PR #176, opened 2018-07-17
Man: pcs.8: primarily refer to the new pacemaker-2.0 daemon man pages

jnpkrn:
Also normalize the relevant punctuation.


tomjelinek:
Thanks for the patch.

Issue #190, opened 2019-01-11
Zombie ruby process with pcs 0.10.1


vvidic:
Starting a pcsd daemon seems to always produce a ruby zombie process:

```
15117 ?        Ss     0:00 /usr/bin/python3 -Es /usr/sbin/pcsd
15121 ?        Z      0:00  \_ [ruby] <defunct>
```

Strace of the parent process suggests it is not waiting for the child to finish?

```
read(11, "{\"next\":1547199195,\"logs\":[]}", 65536) = 29
read(11, 0x123a8e0, 65536)              = -1 EAGAIN (Resource temporarily unavailable)
epoll_wait(6, [{EPOLLHUP, {u32=13, u64=140088948293645}}], 5, -1000) = 1
read(13, "", 65536)                     = 0
sendto(8, "\0", 1, 0, NULL, 0)          = 1
epoll_ctl(6, EPOLL_CTL_DEL, 13, 0x7ffd4452f584) = 0
close(13)                               = 0
epoll_wait(6, [{EPOLLIN, {u32=7, u64=140088948293639}}], 4, 0) = 1
recvfrom(7, "\0", 4096, 0, NULL, NULL)  = 1
recvfrom(7, 0x123a930, 4096, 0, NULL, NULL) = -1 EAGAIN (Resource temporarily unavailable)
epoll_wait(6, [], 4, 0)                 = 0
epoll_wait(6, [{EPOLLHUP, {u32=11, u64=140088948293643}}], 4, -1000) = 1
read(11, "", 65536)                     = 0
sendto(8, "\0", 1, 0, NULL, 0)          = 1
epoll_ctl(6, EPOLL_CTL_DEL, 11, 0x7ffd4452f584) = 0
close(11)                               = 0
epoll_wait(6, [{EPOLLIN, {u32=7, u64=140088948293639}}], 3, 0) = 1
recvfrom(7, "\0", 4096, 0, NULL, NULL)  = 1
recvfrom(7, 0x123ab90, 4096, 0, NULL, NULL) = -1 EAGAIN (Resource temporarily unavailable)
epoll_wait(6, [], 3, 0)                 = 0
epoll_wait(6, [], 3, 0)                 = 0
wait4(9118, 0x7ffd4453020c, WNOHANG, NULL) = 0
epoll_wait(6, [], 3, 0)                 = 0
getpid()                                = 9114
getpid()                                = 9114
epoll_wait(6, [], 3, 0)                 = 0
[...]

PR #191, opened 2019-01-11
avoid zombies by waiting for ruby child exit #190

vvidic:


idevat:
Merged as 3a9901a4fd3cb475ee2235e6df02c6f521857f29. Thanks for the patch.

PR #201, opened 2019-04-12
use more reliable/effortful way of starting systemd units

jnpkrn:
Related to
https://github.com/corosync/corosync/commit/03d9321bc80887d4578744c26c05d61e2d9d4278

### DIFF
--- a/pcs/lib/external.py
+++ b/pcs/lib/external.py
@@ -111,9 +111,18 @@ def start_service(runner, service, instance=None):
         If None no instance name will be used.
     """
     if is_systemctl():
+        # this is the most reliable way to start a service even if it specifies
+        # StopWhenUnneeded=yes
+        # (it's expected the caller of this does require it up at all costs)
         stdout, stderr, retval = runner.run([
-            _systemctl, "start", _get_service_name(service, instance)
+            _systemctl, "enable --runtime --now",
+            _get_service_name(service, instance)
         ])
+        if retval != 0:
+            # systemctl possibly doesn't understand --runtime or --now yet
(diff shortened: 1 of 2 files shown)


tomjelinek:
This change is one part of a fix whose other part is the mentioned corosync/corosync@03d9321. The idea of the corosync part is to stop corosync when it is not used by any other service (in our case, pacemaker).

When pcs starts a cluster, it first starts corosync on all nodes. Once corosync has started, pcs proceeds with starting pacemaker on all nodes. That means corosync is running without any other service using it for a brief time.

That would result in pcs being unable to start a cluster properly: once corosync is started by pcs, systemd shuts it down immediately as pacemaker has not been started yet. This very pcs pull request fixes that: `systemctl enable --now --runtime corosync` enables the corosync service by creating a symlink in `/run`. The symlink prevents systemd from stopping corosync when pacemaker is not running. That is contradicting the original purpose of the fix: stop corosync when pacemaker is not running.

There may be a way to set everything in a way corosync stops when pacemaker is not running and achieve the original intent. The solution, however, adds so much complexity it is not worth implementing.

jnpkrn:
Yes, that was yesterday's conclusion that the usefulness/complexity ratio
make is hardly an appealing proposition.  Thanks for casting other
perspectives to the considerations.

Let me hence retarget the intention here -- from active (wannabe)
corrective actions to passive only dection.  There may also be some
other contraindications hidden in a rich configuration space of
systemd (presets, masks, etc.), but they are not expected to expose
such an internal disconnect (indicating success of the operation
vs. absenting intended effect), so just in case.

PR #251, opened 2020-07-24
Discard master-slave language

bopopescu:
For diversity reasons, it would be nice to try to avoid 'master' and 'slave' terminology in this repository which can be associated to slavery. The master-slave terminology could be problematic for people in several countries which has the history of slavery like Romania, USA and many others. Thank you for considering the proposal. Let me know if any changes in the PR are needed, I would be happy to implement them.


feist:
First, let me assure you this is an issue that is important to this community and has been in progress for some time:
https://lists.clusterlabs.org/pipermail/users/2018-January/014369.html

However, your pull request cannot be accepted for the following reasons:
* It does not pass our regression tests
* It breaks compatibility with other HA cluster components.
* It breaks backward compatibility with older versions of the HA cluster components.
* It breaks compatibility with existing HA cluster configurations deployed at users and customers.
* The changes made to pcs (python) are not reflected in pcsd (ruby) and web UI (js) and are therefore very likely to break even more things.

If this patch can be updated to fix the above issues then we will be happy to consider it for inclusion.

PR #265, opened 2020-11-27
Switch to autotools

fabbione:
Currently missing:
- spec file generation
- gitlab CI integration (we will be able to drop this one once upstream CI is configured properly)
- real life testing (tho all tests are currently passing, including tier1)


fabbione:
Hey guys,

it´s time to start the review and testing.

5 things are missing:
- python embedded modules and gem license files from rpm
- rpm build across all distro (tested only on f33 at the moment)
- testing of the resulting code on a live cluster (tho tier0 and tier1 testing are all passing)
- devel tests are showing some errors, but I am not sure how to fix them
- merge, release, profit!

fabbione:
Hey guys,

one more update compared to the last comment.

Rebased on top of master and solved a couple of minor merge conflicts.
Added automatic generation of file lists for licenses and docs from embedded modules into rpm.

4 things are missing:

* rpm build across all distro (tested only on f33 at the moment)
* testing of the resulting code on a live cluster (tho tier0 and tier1 testing are all passing)
* devel tests are showing some errors, but I am not sure how to fix them
* merge, release, profit!

I am going to tackle the rpm build across all distros today.

fabbione:
Dropping the WIP tag.

CI is green for both normal testing and rpm build targets on all nodes.
Normal testing (tier0): https://ci.kronosnet.org/view/pcs/job/pcs-manual-build-all/65/
rpm: https://ci.kronosnet.org/view/pcs/job/pcs-build-rpms/201/

missing:
* testing of the resulting code on a live cluster (tho tier0 and tier1 testing are all passing)
* devel tests are showing some errors, but I am not sure how to fix them
* merge, release, profit!

Makefile.am
```diff
@@ -0,0 +1,369 @@
...
+	echo 'BUNDLE_BUILD: \"$(ruby_LIBS)\"' >> .bundle/config
+if ENABLE_DOWNLOAD
+else
+	echo 'BUNDLE_ALLOW_OFFLINE_INSTALL: "true"' >> .bundle/config
+endif
+	$(BUNDLE)
+	rm -rf .bundle Gemfile.lock
+endif
+	touch $@
+endif
```
ondrejmular, line 116: Shouldn't `touch $@` be outside of all if blocks?

Makefile.am
```diff
@@ -0,0 +1,369 @@
...
+	touch $@
+
+install-exec-local: install_python_embedded_mods stamps/install_ruby_deps_local
+	$(MKDIR_P) $(or ${DESTDIR}, /)/$(PYTHON_SITELIB)/pcs/
+	$(PYTHON) setup.py install --root=$(or ${DESTDIR}, /) ${EXTRA_SETUP_OPTS} --record $(or ${DESTDIR}, /)/$(PYTHON_SITELIB)/pcs/deinstall.txt
+	$(MKDIR_P) ${DESTDIR}/$(LIB_DIR)/pcs
+	mv ${DESTDIR}/$(SBINDIR)/pcs_internal ${DESTDIR}/$(LIB_DIR)/pcs/pcs_internal
+	mv ${DESTDIR}/$(SBINDIR)/pcs_snmp_agent ${DESTDIR}/$(LIB_DIR)/pcs/pcs_snmp_agent
+	$(PYTHON) -m compileall -fl $(or ${DESTDIR}, /)/$(PYTHON_SITELIB)/pcs/settings.py
```
ondrejmular, line 130: Why is it needed? Is there something special about `pcs/settigns.py` that we need to recompile it?

pcs_test/Makefile.am
```diff
@@ -0,0 +1,386 @@
...
+		$(MKDIR_P) ${builddir}/tools/bin_mock/pcmk/crm_resource.d && \
+		rm -f $${dst} && \
+		$(LN_S) $${src} $${dst}; \
+	done
+
+clean-local:
+	for i in $(AGENTS); do \
+		dst=`echo $${i} | sed -e 's#__#:#g'` && \
+		rm -f $${dst}; \
+	done
```
ondrejmular, line 386: We can avoid this by changing the script a little bit. Here I made that change: https://github.com/ondrejmular/pcs/commit/a33c2a5ae7b0b4bdd077f80ed6dcdc21ad94f9a2

ondrejmular (changes requested):
First of all thank you for all your effort and time that you put into this. You've done a great job which will help us a lot.

I was able to briefly go through the PR, but I'd still like to go into more detail. So far I focused mainly on how all of this will affect our existing workflow. I'll try to briefly sum up my findings here.

### *.in files
#### Python and ruby files
The main issue I see are `*.py.in` and `*.rb.in` files. This greatly affects our workflow in a negative way. These files make the majority of our development tooling ineffective/unusable. The same applies for various editors and python/ruby integrations in them. Luckily, there is an easy workaround for this. AFAIK only paths are replaced in the files, therefore we can move all of them to `settings.py` and `settings.rb` respectively. I'll do all the changes.

#### Executables
There are also some executable files with a template in shabang. For example `#!/usr/bin/env @PYTHON_ENV@`. Wouldn't python3 be sufficient as `env` is already used? This isn't a big deal for me as we can still minimize executable files and move the majority of python code to a separate python file and just import it from the executable.

#### Changelog
This change makes `CHANGELOG.md.in` file not render on github. And we will still need to bump version in it manually so I see no benefit making it `*.in` file. I believe we can stick with the current approach and use `[Unreleased]` for changes not yet included in a release.

### Tarballs from github are unusable
There is no `.version` file in them nor an annotated tag. Maybe we can have the version specified in the repo and we will bump the version together with bumping changelog.

### List of files in Makefile.am
Is it possible to avoid this? Pcs just have too many files to keep lists up to date. I'm sure there will be a ton of cases of us forgetting to update lists. Can we use a wildcard instead of a full list of files?

### Dev tests tools
This approach works for running pylint, mypy and black on the whole project from make. But we are using these tools in editors for realtime checks as well. Editor integration with the tools relies on the tools to be installed so bundling them doesn't work. We can install them separately, but then there is a potential problem of using different versions in the editor and in `make check`.

Therefore I think that we should make use of python virtual env here. This way we can have all dependencies specified in a `dev_requirements.txt` file and install them using `pip` inside a virtual environment. Another improvement will be that there will be no need to take care of dependencies of tools we are using (unlike now where they are listed in `configure.ac`). 

I haven't had a chance to test this approach yet, but I'll do it soon. Do you think that running dev tests in a python virtual environment will be possible in CI?

Makefile.am
fabbione: It doesn´t really matter. it can be inside or outside. The position is purely cosmetic.

ondrejmular:
One last thing that I forgot to mention in previous comment. We can drop support for Travis CI as it doesn't work anymore because we are not eligible for a free plan. That means that whole `test/` directory can be removed together with `.travis.yml` file.

Makefile.am
fabbione: This I don´t know, I carried it over from the old build system due to me not being 100% familiar with python.

pcs_test/Makefile.am
fabbione: Feel free to push those changes directly in the branch, I saw the commit and it´s fine with me of course. The main problem is that autotools in some versions have problems to deal with files that contain ":". Some older version don´t. I have tried escaping in different ways and failed miserably.
Not knowing how the code read the files, I used the hammer approach to rename and symlinks.

fabbione:
> 
> 
> One last thing that I forgot to mention in previous comment. We can drop support for Travis CI as it doesn't work anymore because we are not eligible for a free plan. That means that whole `test/` directory can be removed together with `.travis.yml` file.

Kill it :)

fabbione:
> 
> 
> First of all thank you for all your effort and time that you put into this. You've done a great job which will help us a lot.
> 

thanks

> I was able to briefly go through the PR, but I'd still like to go into more detail. So far I focused mainly on how all of this will affect our existing workflow. I'll try to briefly sum up my findings here.
> ### *.in files
> #### Python and ruby files
> 
> The main issue I see are `*.py.in` and `*.rb.in` files. This greatly affects our workflow in a negative way. These files make the majority of our development tooling ineffective/unusable. The same applies for various editors and python/ruby integrations in them. Luckily, there is an easy workaround for this. AFAIK only paths are replaced in the files, therefore we can move all of them to `settings.py` and `settings.rb` respectively. I'll do all the changes.

That is correct, there are only paths that are affected and I had a brief conversation with Tomas about it and he also suggested that you guys would move those back to use settings.py without substitutions. He mentioned that it would have been too complex for me to do, not knowing why and how some files are used.

Remember that to move them back you also need to update configure.ac and the corresponding Makefile.am.

> #### Executables
> 
> There are also some executable files with a template in shabang. For example `#!/usr/bin/env @PYTHON_ENV@`. Wouldn't python3 be sufficient as `env` is already used? This isn't a big deal for me as we can still minimize executable files and move the majority of python code to a separate python file and just import it from the executable.

I had issues to run the test suite with venv without using env python_version. That´s what sparkled the patch in the first place.
Also some distributions use python and others python3, so it needs to be replaced.

> #### Changelog
> 
> This change makes `CHANGELOG.md.in` file not render on github. And we will still need to bump version in it manually so I see no benefit making it `*.in` file. I believe we can stick with the current approach and use `[Unreleased]` for changes not yet included in a release.

ack and wfm.

> ### Tarballs from github are unusable
> 
> There is no `.version` file in them nor an annotated tag. Maybe we can have the version specified in the repo and we will bump the version together with bumping changelog.

My bad, I forgot to add a couple of files in the tree that handle that specifically. We already solved that problem in knet/corosync/etc.

Let´s finish the rest of the review and I will fix this. I have some pending work that I need to complete.

> ### List of files in Makefile.am
> 
> Is it possible to avoid this? Pcs just have too many files to keep lists up to date. I'm sure there will be a ton of cases of us forgetting to update lists. Can we use a wildcard instead of a full list of files?

Unfortunately no. autotools don´t support wildcards.

> ### Dev tests tools
> 
> This approach works for running pylint, mypy and black on the whole project from make. But we are using these tools in editors for realtime checks as well. Editor integration with the tools relies on the tools to be installed so bundling them doesn't work. We can install them separately, but then there is a potential problem of using different versions in the editor and in `make check`.
> 
> Therefore I think that we should make use of python virtual env here. This way we can have all dependencies specified in a `dev_requirements.txt` file and install them using `pip` inside a virtual environment. Another improvement will be that there will be no need to take care of dependencies of tools we are using (unlike now where they are listed in `configure.ac`).
> 
> I haven't had a chance to test this approach yet, but I'll do it soon. Do you think that running dev tests in a python virtual environment will be possible in CI?

I did use virtual env for a while in CI and now removed after we agreed to go with all embedded or nothing. We have been swinging back and forth a bit on how to handle modules (all in, partially in, etc).

To be honest, we have some level of freedom for devtools. We can drop them from configure.ac or we can make them optional if --enable-devtests is set.

We can also generate requirements.txt based on configure.ac  (so we only need to keep the list in one place) and by default don´t embed the modules.

In general, we can do whatever we want as long as we keep the list of modules in one place (that I would much rather be configure.ac).

fabbione:
For the record, the last commit https://github.com/ClusterLabs/pcs/pull/265/commits/c4190aff8c0674e7ac56c1873a1f022139273b13 adds the ability to download release tarballs directly from github. It is not possible to download snapshots tho. Who wants snapshots might as well use git.

ondrejmular:
I've made some changes to it so it is a little bit more development friendly. @fabbione, can you please take a look at those changes and let me know if it works for you? From our point of view, this PR is ready to be merged so it is not blocking other changes we are working on. Though, there are still some changes needed, mainly with regards to Fedora and RHEL downstream builds which require more flexible rubygems bundling.

Makefile.am
ondrejmular: I left it there for now. I'll get back to it once this PR is merged as it does not affect anything.

ondrejmular (approved):

fabbione:
@ondrejmular all the changes looks ok to me.

I am currently running CI jobs manually to catch any possible fallout and will let you know. It might take a few hours to run them all.

fabbione:
@ondrejmular CI is failing tier1 testing across the board. The error is the same on all nodes. Here is the link to one of them:

https://ci.kronosnet.org/view/pcs/job/pcs-manual-build-all/pcs-manual-build-all=fedora33-x86-64/lastBuild/consoleFull

fabbione:
@ondrejmular as mentioned on IRC last Friday, a test run on CI was good.
Feel free to merge and I will start creating the CI jobs during this week.

PR #272, opened 2021-04-15
require correct tornado version

tomjelinek:
See b75d60054ea9eb898849cc39c15f93a10725be27
and dc345ec159bb7fd601b729f3af314825e75b8939

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -298,7 +298,7 @@ PCS_CHECK_PYMOD([pyOpenSSL])
 # those are kind of problematic.
 # use them all from the BaseOS or embedded them all as necessary (--enable-local-build)
 PCS_CHECK_PYMOD([dacite], [], [yes])
-PCS_CHECK_PYMOD([tornado], [>= 6.0.0], [yes])
+PCS_CHECK_PYMOD([tornado], [>= 6.1], [yes])
 PCS_CHECK_PYMOD([python-dateutil], [>= 2.7.0], [yes])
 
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -298,7 +298,7 @@ PCS_CHECK_PYMOD([pyOpenSSL])
 # those are kind of problematic.
 # use them all from the BaseOS or embedded them all as necessary (--enable-local-build)
 PCS_CHECK_PYMOD([dacite], [], [yes])
-PCS_CHECK_PYMOD([tornado], [>= 6.0.0], [yes])
+PCS_CHECK_PYMOD([tornado], [>= 6.1], [yes])
```
fabbione, line 301: @tomjelinek it fails CI for latest autotools and stuff. You want that to be 6.1.0 (see https://ci.kronosnet.org/job/pcs-build-all-voting/pcs-build-all-voting=ubuntu-20-10-x86-64/lastBuild/console) and there is an unrelated failure here:

https://ci.kronosnet.org/job/pcs-build-all-voting/pcs-build-all-voting=rhel83z-s390x/lastBuild/console

fabbione (changes requested):

configure.ac
fabbione: Apology, I misread the log. Ubuntu-20-10 ships only 6.0.4. If 6.1 is a requirement now, then we need to drop support for Ubuntu.

tomjelinek:
Pcs works with 6.0. I'm cancelling the PR.

PR #282, opened 2021-04-29
fix time-dependant tests

tomjelinek:
Use a longer timeout to make tests less likely to fail
on heavily loaded machines.


tomjelinek:
retest this please

PR #306, opened 2021-06-25
add user friendly error in resource group add

CtrlZmaster:
This patch adds detection of a corner case in 'pcs resource group add'
command and prints a user firendly error. Upon moving this command
to lib, reference checking when deleting a resource group was not moved.
If all of the resources from a group (or cloned group) are deleted,
and there are references from other places in CIB to the deleted group
(or the parent clone), an invalid CIB is produced. Invalid CIB was
dumped originally, this error is now detected and suggests using
the 'pcs resource ungroup' which is implemented in old code and
removes the problematic references.


ondrejmular:
add to whitelist

ondrejmular (approved):

PR #296, opened 2021-05-26
Support starting corosync-notifyd at cluster startup (draft)

inouekazu:
This is a draft of a patch to allow corosync-notifyd to also be started with ```pcs cluster start```.
* ```pcs cluster enable-corosync-notifyd``` will set ```ENABLE_COROSYNC_NOTIFYD=yes``` in /etc/sysconfig/pcs.
  ```
  # pcs cluster enable-corosync-notifyd
  Distributing PCS config...
  rhel84-1: PCS config saved
  rhel84-2: PCS config saved
  ```
  ```
  # ssh rhel84-1 cat /etc/sysconfig/pcs
  # This file has been generated by pcs.
  ENABLE_COROSYNC_NOTIFYD=yes

  # ssh rhel84-2 cat /etc/sysconfig/pcs
  # This file has been generated by pcs.
  ENABLE_COROSYNC_NOTIFYD=yes
  ```
* If this setting is present, corosync-notifyd will also be invoked when the cluster is started.
  ```
  # pcs cluster start --all
  rhel84-1: Starting Cluster...
  rhel84-2: Starting Cluster...

  # ssh rhel84-1 ps -ef | grep -e corosync -e pacemaker
  root        7491       1  5 14:50 ?        00:00:02 /usr/sbin/corosync -f
  root        7504       1  0 14:50 ?        00:00:00 /usr/sbin/corosync-notifyd -f -l -s -m 192.168.122.160
  root        7506       1  0 14:50 ?        00:00:00 /usr/sbin/pacemakerd -f
  haclust+    7507    7506  0 14:50 ?        00:00:00 /usr/libexec/pacemaker/pacemaker-based
  root        7508    7506  0 14:50 ?        00:00:00 /usr/libexec/pacemaker/pacemaker-fenced
  root        7509    7506  0 14:50 ?        00:00:00 /usr/libexec/pacemaker/pacemaker-execd
  haclust+    7510    7506  0 14:50 ?        00:00:00 /usr/libexec/pacemaker/pacemaker-attrd
  haclust+    7511    7506  0 14:50 ?        00:00:00 /usr/libexec/pacemaker/pacemaker-schedulerd
  haclust+    7512    7506  0 14:50 ?        00:00:00 /usr/libexec/pacemaker/pacemaker-controld

  # ssh rhel84-2 ps -ef | grep -e corosync -e pacemaker
  root        6778       1  1 14:50 ?        00:00:03 /usr/sbin/corosync -f
[...]

PR #311, opened 2021-07-01
Support starting and stopping corosync-notifyd

inouekazu:
This patch allows corosync-notifyd to be started with ```pcs cluster start```.
Addressed issues pointed out at https://github.com/ClusterLabs/pcs/pull/296#issuecomment-858555922.


tomjelinek:
Thank you for the patch. We are quite busy right now, but we will get back to you and take a closer look at the patch hopefully in a few weeks.

PR #355, opened 2021-08-13
fix traceback when --wait without timeout is used

ondrejmular:


pcs/lib/env.py
```diff
@@ -176,20 +197,6 @@ def cib(self):
...
-
-        if wait not in self.__timeout_cache:
-            if not self.is_cib_live:
-                raise LibraryError(
-                    ReportItem.error(
-                        reports.messages.WaitForIdleNotLiveCluster()
-                    )
-                )
-            self.__timeout_cache[wait] = get_valid_timeout_seconds(wait)
-        return self.__timeout_cache[wait]
```
tomjelinek, line 191: `__timeout_cache` doesn't seem to be used anywhere, so it can be dropped from `__init__` as well

pcs_test/tools/command_env/config_env.py
```diff
@@ -82,7 +82,7 @@ def push_cib(
         modifiers=None,
         name="env.push_cib",
         load_key="runner.cib.load",
-        wait=False,
+        wait=-1,
```
tomjelinek, line 85: It seems you forgot to update the type in doc text.

tomjelinek (changes requested):
Apart from these small issues it looks good to me.

PR #412, opened 2021-11-25
0.11 release related changes

ondrejmular:


tomjelinek (changes requested):
* `# javascript bundled libraries for old web-ui` sections (there are two) can be removed from spec file, right?
* `Requires: ruby >= 2.2.0` in spec file should be updated.
* `Requires: pacemaker >= 2.0.0` Do we want to change this to 2.1.0?

tomjelinek (approved):

PR #427, opened 2021-12-08
pcsd: fix pcs commands assembly [pcs-0.10]

ondrejmular:


tomjelinek (approved):
I readded 'cluster pcsd-status' command in `app.py`.

PR #430, opened 2021-12-13
revert of disallowing to clone a group with a stonith inside

ondrejmular:
Originaly, this was not fixed properly (it was possible to add a stonith
into a cloned group), therefore to stay consistent, this change is being
reverted. It will be fixed in the future by not allowing stonith to be
placed into a group.

### DIFF
--- a/pcs_test/tier1/cib_resource/test_clone_unclone.py
+++ b/pcs_test/tier1/cib_resource/test_clone_unclone.py
@@ -354,12 +354,9 @@ class Clone(
 
     def test_clone_group_with_stonith(self):
         self.set_cib_file(FIXTURE_GROUP_WITH_STONITH)
-        self.assert_pcs_fail(
+        self.assert_effect(
             "resource clone Group".split(),
-            fixture_clone_stonith_msg(group=True),
-        )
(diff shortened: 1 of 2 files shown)


pcs_test/tier1/cib_resource/test_clone_unclone.py
```diff
@@ -354,20 +354,16 @@ def test_clone_id_is_stonith_forced(self):
...
+            fixture_resources_xml(FIXTURE_CLONED_GROUP_WITH_STONITH),
         )
 
     def test_clone_group_with_stonith_forced(self):
         self.set_cib_file(FIXTURE_GROUP_WITH_STONITH)
         self.assert_effect(
             "resource clone Group --force".split(),
             fixture_resources_xml(FIXTURE_CLONED_GROUP_WITH_STONITH),
-            output=fixture_clone_stonith_msg(forced=True, group=True),
         )
```
tomjelinek, line 367: That's the same test twice.

tomjelinek (changes requested):

tomjelinek (approved):

PR #431, opened 2021-12-15
Fix snmp client

vvidic:
Required constant is missing causing the command to fail on startup and breaking the pcs_snmp_agent service.


tomjelinek:
test this please

tomjelinek (approved):

tomjelinek:
Thank you for the patch!

PR #452, opened 2022-01-18
fix rhbz#1954099

F16shen:
Fix rhbz#1954099:https://bugzilla.redhat.com/show_bug.cgi?id=1954099
A patch to fix preventing fence-loop caused when stonith-watchdog-timeout is set with wrong value


tomjelinek:
test this please

F16shen:
> test this please

Could someone review this please?

tomjelinek (changes requested):
Hi @F16shen,

Thank you pro sending this patch and sorry for the delay. I went through the patch and found several issues which we would like to get fixed before merging it.

There is no need for `verify_stonith_watchdog_timeout` function to be called for all properties as it is only meant for processing `stonith-watchdog-timeout`. Instead `args`, the parameter should be the property's value. The function should not be responsible for calling `service_manager`, it should get a boolean parameter describing whether sbd is enabled instead. Similarly with sbd devices and timeout. The function definitely needs to return `pcs.common.reports.ReportItemList` instead of text messages. Actually, it might be best to split this function, put the part which produces messages to properties module and keep the other part(s) in sbd module. Type hints are missing, as well as tests and changelog.

Since this bug is getting near the top of our todo list, we can fix the issues ourselves, if you prefer so. Your commit would be of course included, so you wouldn't lose credits for it.

Regards,
Tomas

F16shen:
@tomjelinek Hi, Tomas, Thank you for the detailed explanation and sorry for the delay. I have fixed some issues that you mentioned before. 

I am not so sure the way I use ```pcs.common.reports.ReportItemList``` is the right way, I think it should have a cleaner way.

Pleas consider when you have time.

tomjelinek:
test this please

tomjelinek:
retest this please

tomjelinek:
retest this please

tomjelinek:
test this please

PR #463, opened 2022-02-03
process invalid OCF agents as if they complied with OCF 1.0 [pcs-0.10]

tomjelinek:


pcs/lib/resource_agent/xml.py
```diff
@@ -137,20 +138,36 @@ def load_fake_agent_metadata(
...
+            reports.ReportItem.warning(
+                reports.messages.AgentImplementsUnsupportedOcfVersionAssumedVersion(
+                    name.full_name,
+                    ocf_version,
+                    sorted(const.SUPPORTED_OCF_VERSIONS),
+                    const.OCF_1_0,
+                )
+            )
+        )
+    return _parse_agent_1_0(name, metadata)
```
ondrejmular, line 170: We should treat case of unsupported ocf version more carefully as there is no prior schema validation. It can end up with some unhandled exceptions caused by accessing non-existing attributes of a xml element. Therefore we should create a separate function similar to `_parse_agent_1_0` where missing attributes which are not vital for pcs to work can be ignored and raise an exception if it's not possible to continue.  Or another option would be to use `_parse_agent_1_0` as it is but catch any possible exception and raise our own exception if there was any problem. Either way, the exception produced by our code should then be handled by higher layers (printing out an error or skipping the agent).

pcs/common/reports/codes.py
```diff
@@ -39,6 +39,9 @@
 AGENT_IMPLEMENTS_UNSUPPORTED_OCF_VERSION = M(
     "AGENT_IMPLEMENTS_UNSUPPORTED_OCF_VERSION"
```
ondrejmular, line 40: I believe this is not used as `UnsupportedOcfVersion` is not raised anymore

pcs/lib/resource_agent/xml.py
```diff
@@ -137,20 +138,36 @@ def load_fake_agent_metadata(
 
 
 def parse_metadata(
-    name: ResourceAgentName, metadata: _Element
+    name: ResourceAgentName,
+    metadata: _Element,
+    report_processor: reports.ReportProcessor,
+    report_warnings=True,
```
ondrejmular, line 144: I don't like the idea of passing 2 extra arguments to this function. This just doesn't seems to be the best place to solve this. I would prefer to keep this low level function simple (not dealing with reports).

As we don't want to always report this warning (I guess it's due to not outputting warnings to stderr), we should handle this case in higher layers. Returning real  ocf version of source metadata as another return value should enable us to make this check somewhere else and only in when its needed.

ondrejmular (changes requested):

pcs/lib/resource_agent/xml.py
tomjelinek: I didn't like it either :)

pcs/lib/resource_agent/facade.py
```diff
@@ -239,6 +244,7 @@ def _get_fenced_parameters(self):
                             load_fake_agent_metadata(
                                 self._runner, agent_name.type
                             ),
+                            self._report_processor,
```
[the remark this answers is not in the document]
tomjelinek: ack

pcs/common/reports/codes.py
tomjelinek: ack, removed

pcs/lib/resource_agent/xml.py
tomjelinek: That's not really worth the effort. The purpose of this patch is to allow valid agents to be processed even if they report bad OCF version (due to version tag and version attribute confusion). Such agents are almost always processable by `_parse_agent_1_0`, which treats majority of elements and attributes as optional anyway. In the rare cases of an exception, we're fine with it. And it's caught in higher layers anyway.

ondrejmular: I understand the purpose of this patch. However, I still think pcs should not exit with a traceback in case agent's version is not supported and there is actually a mistake in agent's metadata (e.g. missing name attribute in parameter xml tag). This is not the case for agents with supported versions as their metadata are validated against xml schema.

I gave it a try and I added a parameter without name in test `pcs_test.tier0.lib.resource_agent.test_facade.ResourceAgentFacadeFactory.test_facade_bad_ocf_version` and it is actually failing with a `KeyError`.

Therefore we should either handle this exception somewhere or modify `pcs.lib.resource_agent.xml._metadata_xml_to_dom` in a way that unsupported metadata versions are also validated  with 1.0 schema (similarly how `pcs.lib.resource_agent.xml.parse_metadata` has been updated).

ondrejmular: Thinking about it little bit more, I can see a downside of ocf 1.0 validation in `pcs.lib.resource_agent.xml._metadata_xml_to_dom` for metadata with an unsupported version. In case it's not valid pcs will only print out error about wrong format of metadata (which is fine because pcs would fail later on anyway) but there will be no warning about unsupported metadata format version. But I believe this is not a big deal.

tomjelinek: I added a bunch of fixes to make pcs process OCF 1.0 agents in the same way as they were processed before resource agent parser overhaul.

pcs/lib/resource_agent/ocf_transform.py
```diff
@@ -170,3 +194,14 @@ def _bool_value(value: Optional[str]) -> bool:
     value -- raw bool value
     """
     return value == "1"
+
+
+def _bool_value_1_0(value: Optional[str]) -> bool:
```
ondrejmular, line 199: Naming of `_bool_value` vs `_bool_value_1_0` could be misleading as `1_0` in the name could mean ocf 1.0 or `1` and `0` as only possible bool values. Therefore I would suggest to rename `_bool_value_1_0` to `_bool_value_ocf_1_0` or to `_bool_value_legacy` as it doesn't completely conform to ocf 1.0.

ondrejmular (approved):

pcs/lib/resource_agent/ocf_transform.py
tomjelinek: ack, changed to `_bool_value_legacy`

PR #464, opened 2022-02-03
fix xpath variables

tomjelinek:


pcs/lib/cib/acl.py
```diff
@@ -325,7 +329,9 @@ def get_role_list(acl_section):
     acl_section -- etree node
     """
     output_list = []
-    for role_el in acl_section.findall("./{0}".format(TAG_ROLE)):
+    for role_el in acl_section.xpath(
+        "./*[local-name()=$tag_name]", tag_name=TAG_ROLE
+    ):
```
ondrejmular, line 334: Here it's not necessary to use xpath variable injection as `TAG_ROLE` is internal constant and not an input from user. In my opinion it makes the code just a little bit more complex and harder to read than necessary. This is also the case in other places.

On the other hand, maybe it's safer to do it for everything without thinking about source of the value. So this is not really a disagreement, just an observation.

pcs/lib/cib/acl.py
```diff
@@ -229,9 +232,10 @@ def create_target(acl_section, target_id):
     """
     # id of element acl_target is not type ID in CIB ACL schema so we don't need
     # to check if it is unique ID in whole CIB
-    if (
-        acl_section.find("./{0}[@id='{1}']".format(TAG_TARGET, target_id))
-        is not None
+    if acl_section.xpath(
+        "./*[local-name()=$target_tag and @id=$target_id]",
+        target_tag=TAG_TARGET,
+        target_id=target_id,
```
ondrejmular, line 238: Based on the comment about inserting constants into a xpath expresion, this could be changed like so:
```suggestion
    if acl_section.xpath(
        f"./{TAG_TARGET}[@id=$target_id]",
        target_id=target_id,
```

pcs/lib/cib/fencing_topology.py
```diff
@@ -425,26 +425,44 @@ def _append_level_element(tree, level, target_type, target_value, devices):
...
     if xpath_attrs:
-        return tree.xpath("fencing-level[{0}]".format(xpath_attrs))
+        return tree.xpath(
+            f"fencing-level[{xpath_attrs}]",
+            var_target=var_target,
+            var_target_pattern=var_target_pattern,
+            var_target_attribute=var_target_attribute,
+            var_target_value=var_target_value,
+            var_devices=(",".join(devices) if devices else ""),
+            var_level=level,
```
ondrejmular, line 466: `dict` can be used instead of separate variables and then it can be expanded here using `**`

pcs/lib/cib/resource/remote_node.py
```diff
@@ -76,18 +76,19 @@ def find_node_resources(resources_section, node_identifier):
...
                         and
-                        @value="{identifier}"
+                        @value=$identifier
                     ]
                 )
             ]
         """.format(
-            is_remote=_IS_REMOTE_AGENT_XPATH_SNIPPET, identifier=node_identifier
-        )
+            is_remote=_IS_REMOTE_AGENT_XPATH_SNIPPET
```
ondrejmular, line 89: Could be replaced by f-string

pcs/lib/cib/tag.py
```diff
@@ -463,7 +463,7 @@ def _find_obj_ref_in_tag(self, obj_ref_id: str) -> Optional[_Element]:
         xpath_result = cast(
             List[_Element],
             self._tag_element.xpath(
-                f'./{TAG_OBJREF}[@id="{obj_ref_id}"]',
+                f"./{TAG_OBJREF}[@id=$obj_ref_id]", obj_ref_id=obj_ref_id
```
ondrejmular, line 466: Here it is used as I suggested in an another comment :)

ondrejmular (approved):

pcs/lib/cib/acl.py
tomjelinek: ack

pcs/lib/cib/acl.py
tomjelinek: It's safer to pass everything as variables. In this very case, however, it's an over-complication.

pcs/lib/cib/resource/remote_node.py
tomjelinek: ack

pcs/lib/cib/fencing_topology.py
tomjelinek: ack

ondrejmular (approved):

PR #475, opened 2022-03-04
Add support for fence_mpath to `pcs stonith update-scsi-devices` command

mirecheck:


pcs_test/tier0/lib/cib/resource/test_stonith.py
```diff
@@ -279,3 +286,140 @@ def test_unsupported_provider(self):
...
+            {"rh9-1": "1", "rh9-2": "2", "rh9-3": "3"},
+        )
+
+    def test_success_mixed_assigment_char(self):
+        self.assert_success(
+            "rh9-1:1;rh9-2=2;rh9-3:3",
+            {"rh9-1": "1", "rh9-2": "2", "rh9-3": "3"},
+        )
+
+    def test_scucess_semicolon_as_separator(self):
```
ondrejmular, line 378:
```suggestion
    def test_success_semicolon_as_separator(self):
```
The same typo is in tests below as well.

ondrejmular (changes requested):

pcs/lib/cib/resource/stonith.py
```diff
@@ -140,6 +145,39 @@ def validate_stonith_restartless_update(
...
+    )
+    pcmk_host_map_value = get_value(
+        INSTANCE_ATTRIBUTES_TAG, stonith_el, "pcmk_host_map"
+    )
+    if not pcmk_host_map_value:
+        raise library_error(
+            "mpath reservation keys not found in 'pcmk_host_map' attribute"
+        )
+    node_key_map = {}
+    pattern = re.compile(r"(?P<node>[^=:; \t]+)[=:](?P<key>[^=:; \t]+)[; \t]?")
```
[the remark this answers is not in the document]
mirecheck: Here is pacemaker function which parses pcmk_host_map value: https://github.com/ClusterLabs/pacemaker/blob/main/daemons/fenced/fenced_commands.c#L677-L755
It is not exactly 1:1 copy because i skipped the skipping of escaped characters.

pcs/common/reports/messages.py
```diff
@@ -2950,6 +2950,36 @@ def message(self) -> str:
...
+        nodes = format_plural(self.missing_nodes, "node")
+        node_list = format_list(self.missing_nodes)
+        node_names = f": {node_list},"
+        if not self.missing_nodes:
+            keys = "keys"
+            nodes = "nodes"
+            node_names = ""
+        return (
+            f"Missing mpath reservation {keys} for {nodes}{node_names} in "
+            f"'pcmk_host_map' value: '{self.pcmk_host_map_value}'"
```
ondrejmular, line 2979: Maybe `pcmk_host_map_value` should be `Optional[str]` and in case its value is `None` the message could look something like this: `Missing mpath reservation keys, 'pcmk_host_map' not set.`

ondrejmular (changes requested):

ondrejmular (approved):

PR #483, opened 2022-03-18
notify user about existing move constraints

mirecheck:
* updated commands:
  * `pcs resource move-with-constraint`
  * `pcs status`


tomjelinek:
retest this please

tomjelinek (changes requested):
The `pcs status` part is ok except those few minor issues.

The `pcs resource move` part is not ok as pcs cannot reliably figure out effect of other constraints on the moved resource. On top of that, pacemaker can actually remove constraints related to a resource on moving it. So, instead of analyzing the situation, we will do this: if `crm_resource --move` returns 0 and no stdout or stderr output, pcs will print a message that a constraint has been created and the resource may or may not move depending on other configuration. This will be only included in `pcs resource move`.

pcs/lib/commands/resource.py
```diff
@@ -1967,6 +1967,10 @@ def _report_action_pcmk_error(self, resource_id, stdout, stderr):
     def _report_action_pcmk_success(self, resource_id, stdout, stderr):
         raise NotImplementedError()
 
+    @staticmethod
+    def _report_resource_may_or_may_not_move(resource_id):
+        raise NotImplementedError()
```
tomjelinek, line 1972:
```suggestion
    def _reports_resource_may_or_may_not_move(resource_id: str) -> reports.ResourceItemList:
        return []
```

pcs/lib/commands/resource.py
```diff
@@ -2025,6 +2030,10 @@ def run(
             raise LibraryError(
                 self._report_action_pcmk_error(resource_id, stdout, stderr)
             )
+        if node and not stdout and not stderr:
+            env.report_processor.report(
+                self._report_resource_may_or_may_not_move(resource_id)
+            )
```
tomjelinek, line 2036:
```suggestion
            env.report_processor.report_list(
                self._reports_resource_may_or_may_not_move(resource_id)
            )
```

pcs/lib/commands/resource.py
```diff
@@ -2076,6 +2085,12 @@ def _report_action_pcmk_success(self, resource_id, stdout, stderr):
             )
         )
 
+    @staticmethod
+    def _report_resource_may_or_may_not_move(resource_id):
+        return ReportItem.warning(
+            reports.messages.ResourceMayOrMayNotMove(resource_id)
+        )
```
tomjelinek, line 2092:
```suggestion
    def _reports_resource_may_or_may_not_move(resource_id: str) -> reports.ReportItemList:
        return [ReportItem.warning(
            reports.messages.ResourceMayOrMayNotMove(resource_id)
        )]
```

pcs/lib/commands/resource.py
```diff
@@ -2122,6 +2137,10 @@ def _report_action_pcmk_success(
             reports.messages.ResourceBanPcmkSuccess(resource_id, stdout, stderr)
         )
 
+    @staticmethod
+    def _report_resource_may_or_may_not_move(resource_id):
+        return []
+
```
tomjelinek, line 2143:
```suggestion
```
You cannot pass a list to `ReportProcessor.report`

pcs_test/tier0/lib/commands/test_status.py
```diff
@@ -867,3 +867,59 @@ def test_daemon_status_all_off(self):
...
+            constraints="""
+            <constraints>
+                <rsc_location id="cli-ban-P1-on-node1" rsc="P1"
+                    role="Started" node="node1" score="-INFINITY"/>
+                <rsc_location id="cli-prefer-P1" rsc="P1" role="Started"
+                    node="node3" score="INFINITY"/>
+                <rsc_location id="cli-prefer-P2" rsc="P2" role="Started"
+                    node="node1" score="INFINITY"/>
+                <rsc_location id="cli-ban-P2-on-node1" rsc="P2"
+                    role="Started" node="node1" score="-INFINITY"/>
```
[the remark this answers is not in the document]
tomjelinek: I'm sorry, how is this resolved? The only non-cli constraint I can see here is for P2 resource and that resource is listed in the warning bellow... Am I missing something?

tomjelinek (changes requested):

tomjelinek (approved):

PR #502, opened 2022-04-25
fix bug 1899519

mengfansheng-git:
fix bug 1899519

### DIFF
--- a/pcs/lib/cib/nvpair.py
+++ b/pcs/lib/cib/nvpair.py
@@ -32,9 +32,27 @@ def _append_new_nvpair(nvset_element, name, value, id_provider):
         "nvpair",
         id=create_subelement_id(nvset_element, name, id_provider),
         name=name,
-        value=value,
+        value=clean_xml_string(value)
     )
 
+def is_valid_xml_char_ordinal(i):
+    """
+    Defines whether char is valid to use in xml document
+    XML standard defines a valid char as::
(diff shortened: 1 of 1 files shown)


tomjelinek:
Thank you very much for the pull request. However, we decided not to accept it in its current form.

The proposed change makes pcs silently drop characters from user input, which leads to saving a different value to CIB than the requested one, all of that without notifying users in any way. Such behavior is very confusing and is considered to be a bug.

The proper way would be to print an error that an entered value is not valid. Even that may no longer be necessary, though, as [bug 1899519](https://bugzilla.redhat.com/show_bug.cgi?id=1899519) seems to only be reproducible with Python 2, which is no longer supported by pcs.

mengfansheng-git:
This bug seems to exist in python3 as well
image

tomjelinek:
Could you provide a reproducer using pcs? We're still investigating if this is a pcs issue, and your screenshot of a traceback doesn't seem to be related.

Your new commit reports an error, that's an improvement in the right direction. But it disallows storing all characters with ord < 128 to meta attribute values. However, the original issue in [bug 1899519](https://bugzilla.redhat.com/show_bug.cgi?id=1899519) states that it is required to store such characters.

mengfansheng-git:
Can add an exception catch, not to the user input the node name is not standardized without prompting directly report an error

PR #512, opened 2022-05-09
fix a bunch of typos

tomjelinek:


pcs/lib/commands/cib_options.py
```diff
@@ -152,7 +152,7 @@ def _defaults_create(
     )
 
     env.report_processor.report(
-        ReportItem.warning(reports.messages.DefaultsCanBeOverriden())
+        ReportItem.warning(reports.messages.DefaultsCanBeOverridden())
```
ondrejmular, line 155: Re report messages: In order to keep codes and message class names consistent, we should create new codes with fixed typos and make sure old codes could not be reused anymore.

ondrejmular (changes requested):

ondrejmular (approved):

PR #518, opened 2022-05-30
pcs resource manage --monitor enables monitor for all resources in a group

F16shen:
Fix rhbz#1918527:https://bugzilla.redhat.com/bugzilla/show_bug.cgi?id=1918527

I use the same code in `find_resources_to_unmanage` to find the specified resource, I'm not sure if this is the right way to do it.

### DIFF
--- a/pcs/lib/cib/resource/common.py
+++ b/pcs/lib/cib/resource/common.py
@@ -273,6 +273,24 @@ def find_resources_to_manage(resource_el):
     )
 
 
+def find_resources_to_manage_with_monitor(resource_el):
+    """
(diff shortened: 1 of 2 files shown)


tomjelinek:
test this please

tomjelinek:
Hi @F16shen,

Thank you for the pull request! We appreciate your effort.

In this case, however, it doesn't seem to me you managed to fix the bug. When I applied your patch and ran a reproducer from [rhbz#1918527](https://bugzilla.redhat.com/bugzilla/show_bug.cgi?id=1918527), I got the same results as without the patch.

Since I already dived into the issue to be able to review your code, I fixed it in #519.

PR #557, opened 2022-09-13
show specific properties which are not explicitly configured

F16shen:
fix rhbz2112267: https://bugzilla.redhat.com/bugzilla/show_bug.cgi?id=2112267

### DIFF
--- a/pcs/resource.py
+++ b/pcs/resource.py
@@ -2483,6 +2483,10 @@ def resource_disable_common(
             modifiers.get("--wait"),
         )
         return
+    if modifiers.get("--brief"):
+        raise CmdLineInputError(
+            "'--brief' cannot be used without '--simulate' or '--safe'"
+        )
     lib.resource.disable(argv, modifiers.get("--wait"))
 
 
--- a/pcs_test/tier0/cli/test_resource.py
+++ b/pcs_test/tier0/cli/test_resource.py
@@ -690,6 +690,20 @@ class ResourceDisable(TestCase):
         self.resource.disable_safe.assert_not_called()
         self.resource.disable_simulate.assert_not_called()
 
+    def test_brief(self):
+        with self.assertRaises(CmdLineInputError) as cm:
+            resource.resource_disable_common(
+                self.lib, ["R1", "R2"], dict_to_modifiers(dict(brief=True))
+            )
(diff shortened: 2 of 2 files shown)


tomjelinek:
test this please

pcs/prop.py
```diff
@@ -147,12 +147,15 @@ def list_property(lib, argv, modifiers):
...
 
+
     if not modifiers.get("--defaults"):
         configured_properties = utils.get_set_properties(
             None if print_all else argv[0], properties
         )
         if modifiers.get("--all"):
             properties.update(configured_properties)
+        elif not configured_properties:
+            properties = get_default_property(argv[0])
```
tomjelinek, line 158: What is the point of loading properties again? They are already loaded in the `properties` variable.

pcs/prop.py
```diff
@@ -161,6 +164,18 @@ def list_property(lib, argv, modifiers):
...
+def get_default_property(property_name):
+    """
+    Commandline options: no options
+    """
+    parameter = {}
+    prop_def_dict = utils.get_cluster_properties_definition()
+    for name, prop in prop_def_dict.items():
+        if name == property_name:
+            parameter[name] = prop["default"]
+    return parameter
```
tomjelinek, line 176: This function is the same as existing function `get_default_properties`, except it filters the result. It's better to use the existing function and filter its result to avoid code duplication.

tomjelinek (changes requested):
Thank you for your patch. Could you resolve the issues found during review, fix the patch to make it pass our test suite, and add a testcase for the new behavior?

F16shen:
Thanks for the reply. 
I have resolved the issues you mentioned before, I also pass the testcase locally with this command:
`python3 pcs_test/suite --tier1 -v --vanilla pcs_test.tier1.legacy.test_properties.PropertyTest`

And perhaps it's better to show a error message like : `Error: unknown cluster property: 'xxxx'` other than `Cluster Properties:` for those who use `pcs property config` on invalid option?

tomjelinek:
Sorry, I closed the PR by mistake and GitHub doesn't allow me to reopen it. I opened #565 instead.

Thanks for the update! Once this passes CI, we'll merge it.

PR #561, opened 2022-09-19
--brief option usage does not match description

F16shen:
Description of problem:
pcs resource disable usage shows --brief option should be use after --safe or --simulate option:
```    
disable <resource id | tag id>... [--safe [--brief] [--no-strict]]
            [--simulate [--brief]] [--wait[=n]]
```

If use --brief option alone, it will also take effect without any output, which is confusing.

Perhaps use warning is better? But in that way I'm not sure how to modify tests.

### DIFF
--- a/pcs/resource.py
+++ b/pcs/resource.py
@@ -2483,6 +2483,11 @@ def resource_disable_common(
             modifiers.get("--wait"),
         )
         return
+    if modifiers.get("--brief"):
+        raise CmdLineInputError(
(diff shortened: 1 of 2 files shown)


tomjelinek:
test this please

tomjelinek:
retest this please

tomjelinek:
retest this please

tomjelinek:
retest this please

Issue #235, opened 2019-12-18
[input validation] pcs shall check that location constraint always relates to some nodes in one way or another


jnpkrn:
```
# pcs constraint location dummy rule id=till-the-new-year score=-INFINITY date gt 2020-01-01 
# echo $?
> 0
```

->

```
      <rsc_location id="location-dummy" rsc="dummy">
        <rule id="till-the-new-year" score="-INFINITY">
          <date_expression id="till-the-new-year-expr" operation="gt" start="2020-01-01"/>
        </rule>
      </rsc_location>
```

This is a degenerate location constraint that achieves exactly nothing
but more complex overall configuration -- there's no actual "location"
specified in there.

At the very least, this shall raise a warning.

Relevant:
https://bugs.clusterlabs.org/show_bug.cgi?id=5415

tomjelinek:
ack, tracked at https://bugzilla.redhat.com/show_bug.cgi?id=1791791

daaitudian:
Thanks very much for your help.

PR #562, opened 2022-09-22
fix location constraint bug 1791791

daaitudian:
Hi:
In this Pr, I made some fixes for the #235


tomjelinek:
test this please

pcs/constraint.py
```diff
@@ -926,9 +926,11 @@ def _show_location_rules(
     return all_lines
 
 
-def _verify_node_name(node, existing_nodes):
+def _verify_node_name(node, existing_nodes, warn: bool = False):
     report_list = []
     if node not in existing_nodes:
+        if warn :
+            return False
```
tomjelinek, line 933: * This function returns a list of ReportItem objects. It must never return `False`.
* Instead of calling a function and passing it an argument to tell the function to do nothing, it is better not to call the function at all.
* The function doesn't seem to be called with `warn` parameter anywhere. What's the parameter purpose than?

pcs/constraint.py
```diff
@@ -1195,6 +1197,19 @@ def location_rule(lib, argv, modifiers):
             "resource-discovery": None,
         },
     )
+
+    if options["node"] == None or not options["node"]:
+        warn("There's no actual 'location' specified in this constraint, you can specify it by node. ")
```
tomjelinek, line 1202: This is a misunderstanding of how location constraints with rules work.

There is no `node` option. The node is specified by the rule itself. For example, rule `node_size gte 42` applies to all nodes which have the value of their attribute `node_size` greater than or equal to `42`.

If, however, there are only date expression in the rule, like shown in #235, then the rule applies to all nodes and is therefore pointless in a location constraint.

You can check `pcs.lib.cib.rule.tools` to see how to analyze a rule to figure out if it contains node expressions. The module, however, is part of pcs library, new architecture, and is not compatible with the rule parser used in location constraints.

pcs/rule.py
```diff
@@ -114,8 +114,8 @@ def dom_rule_add(dom_element, options, rule_argv, cib_schema_version):
         rule.setAttribute("score", "0")
     if dom_element.hasAttribute("score"):
         dom_element.removeAttribute("score")
-    if dom_element.hasAttribute("node"):
-        dom_element.removeAttribute("node")
+    #if dom_element.hasAttribute("node"):
+    #    dom_element.removeAttribute("node")
```
tomjelinek, line 118: I'm not sure what is the point of this change.

pcs/rule.py
```diff
@@ -22,7 +22,7 @@ def parse_argv(argv, extra_options=None):
     """
     Commandline options: no options
     """
-    options = {"id": None, "role": None, "score": None, "score-attribute": None}
+    options = {"id": None, "role": None, "node":None, "score": None, "score-attribute": None}
```
tomjelinek, line 25: I'm not sure what is the point of this change. As explained in the other comment, there is no `node` option in location constraints.

tomjelinek (changes requested):
We appreciate the effort you put into this patch. However, we cannot accept it in its current form.

daaitudian:
Thank for your reply. When I see this issue, I thought it was mean to add a [node] option for rule command to make sure  the specific resource can run in the node under the rule.
Since [node] option is not required in this command, so how would you like to fix this issue?

tomjelinek:
> how would you like to fix this issue?

The correct way is to analyze the constraint rule and check if it contains any node expressions (`defined`, `not_defined`, `lt`, `gt`, `lte`, `gte`, `eq`, `ne`). If it doesn't, print a warning. The check cannot be implemented simply, e.g. `"lt" in rule`. It needs to understand the rule structure to avoid cases when `lt` is used in a date expression or not as an operator.

daaitudian:
Ok, thanks for your explanation. I will modify it later.

Issue #552, opened 2022-08-25
Do not use unmaintained `open4` gem in `pcsd`


lucaskanashiro:
The `open4` gem is one of the dependencies of `pcsd`, however, it is unmaintained for a while ([last activity in its git repository dated to 2014](https://github.com/ahoward/open4)). It would be great if we could move to use a gem not so outdated, like [posix-spawn](https://github.com/rtomayko/posix-spawn) or [childprocess](https://github.com/enkessler/childprocess).

PR #568, opened 2022-10-05
pcsd: use childprocess gem instead of open4 (Closes: #552)

lucaskanashiro:
The open4 gem is unmaintained upstream for a while, the last activity in its git repository is from 2014. On the other hand, the childprocess maintenance is more active recently, and it also used by some other projects written in ruby, like vagrant and selenium-webdriver.


lucaskanashiro:
FWIW I tested this patch in the test suite we use in Ubuntu and all tests passed just fine. Here is the test suite:

https://github.com/canonical/server-test-scripts/tree/main/ha/virsh

ondrejmular:
test this please

lucaskanashiro:
Thanks for triggering the tests @ondrejmular . I noticed the failures in `debian-testing`, `debian11` and `ubuntu-22-04-lts`, all of them are failing because of this:

```
configure: error: ruby gem childprocess not found
```

With the proposed patch, `childprocess` needs to be installed during configure time (instead of `open4`). I am wondering where is the code executed in your CI system, because it is installing `open4` instead of `childprocess`, and there is no occurrence of `open4` in the proposed branch. My guess is that this is hosted somewhere else and it needs update. Could you please clarify that to me?

tomjelinek (approved):
Thank you very much for this patch. It's just perfect. Once the CI issue is resolved, we're going to merge it.

lucaskanashiro:
Thank you very much @tomjelinek ! Waiting for that :)

tomjelinek:
retest this please

tomjelinek:
retest this please

tomjelinek:
retest this please

PR #573, opened 2022-10-17
More precise error messages for moving unmanaged resource

F16shen:
In my case, I don't have any constraint in the cluster , so the error message looks quite confusing.
```
Error: Unable to move resource 'xxxxx' using a location constraint. Current location of the resource may be affected by some other constraint.
```

Does this needs a test case?

### DIFF
--- a/pcs/lib/commands/resource.py
+++ b/pcs/lib/commands/resource.py
@@ -1698,8 +1698,9 @@ def move_autoclean(
     if env.report_processor.report_list(report_list).has_errors:
         raise LibraryError()
 
+    cluster_state=env.get_cluster_state()
     resource_state_before = get_resource_state(
-        env.get_cluster_state(), resource_id
+        cluster_state, resource_id
     )
     if not _resource_running_on_nodes(resource_state_before):
         raise LibraryError(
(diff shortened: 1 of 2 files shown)


tomjelinek:
test this please

pcs/lib/commands/resource.py
```diff
@@ -1698,15 +1698,23 @@ def move_autoclean(
...
     )
     if not _resource_running_on_nodes(resource_state_before):
         raise LibraryError(
             ReportItem.error(
                 reports.messages.CannotMoveResourceNotRunning(resource_id)
             )
         )
+    if not is_resource_managed(cluster_state, resource_id):
+        raise LibraryError(
+            ReportItem.warning(
```
ondrejmular, line 1713: This should be an error as we are basically sure move will fail.

pcs/lib/commands/resource.py
```diff
@@ -1698,15 +1698,23 @@ def move_autoclean(
     if env.report_processor.report_list(report_list).has_errors:
         raise LibraryError()
 
+    cluster_state=env.get_cluster_state()
```
ondrejmular, line 1701:
```suggestion
    cluster_state = env.get_cluster_state()
```

We use https://github.com/psf/black for python code formatting. Make sure that you run `make check` before committing changes. To enable all development related checks (e.g. pylint, mypy, black and typos checks) run `./configure` with option `--enable-dev-tests`. See [CONTRIBUTING.md](https://github.com/ClusterLabs/pcs/blob/main/CONTRIBUTING.md) for more details.

You can use `make black` to automatically format all python code.

ondrejmular (changes requested):
Thank you for the patch. I like the idea.

Could you please also add a test coverage for the new code? Adding a test-case to `pcs_test.tier0.lib.commands.resource.test_resource_move_autoclean.MoveAutocleanValidations` should be sufficient.

F16shen:
Thank you for the reply. I have fixed issue you mentioned @ondrejmular

ondrejmular (approved):

tomjelinek:
retest this please

PR #598, opened 2022-11-28
Support ISO 8601 duration specifications for pacemaker "time" properties

daaitudian:
Hi：
This PR is uesd to support ISO 8601 duration specifications.
 [https://bugzilla.redhat.com/show_bug.cgi?id=1807636](url)

### DIFF
--- a/pcs_test/tier0/lib/commands/test_cluster_property.py
+++ b/pcs_test/tier0/lib/commands/test_cluster_property.py
@@ -911,6 +911,10 @@ class TestGetProperties(TestCase):
         )
         self.env_assist.assert_reports([])
 
+    @mock.patch(
+        "pcs.lib.cib.rule.in_effect.has_rule_in_effect_status_tool",
+        lambda: True,
+    )
     def test_evaluate_expired_but_no_set_rule(self):
         self.config.runner.cib.load(
             crm_config=fixture_crm_config_properties([("set_id", {})])
@@ -922,6 +926,30 @@ class TestGetProperties(TestCase):
                     CibNvsetDto(id="set_id", options={}, rule=None, nvpairs=[])
                 ]
             ),
+        )
+
+    @mock.patch(
+        "pcs.lib.cib.rule.in_effect.has_rule_in_effect_status_tool",
+        lambda: False,
+    )
+    def test_evaluate_expired_no_status_tool(self):
+        self.config.runner.cib.load(
+            crm_config=fixture_crm_config_properties([("set_id", {})])
+        )
+        self.assertEqual(
+            self.command(evaluate_expired=True),
+            ListCibNvsetDto(
+                nvsets=[
+                    CibNvsetDto(id="set_id", options={}, rule=None, nvpairs=[])
+                ]
+            ),
+        )
+        self.env_assist.assert_reports(
+            [
+                fixture.warn(
+                    reports.codes.RULE_IN_EFFECT_STATUS_DETECTION_NOT_SUPPORTED,
+                )
+            ]
         )
 
 


tomjelinek:
test this please

daaitudian:
Hi：
I see that this pr has not been passed, and I want to know how I can modify the current problem. Looking forward to your reply！

mirecheck:
Hi @daaitudian,

First, I would like to thank you for your pull request. I’m sorry that it took so long to look at your pull request. We were overhauling all cluster properties code, and that’s the reason it took us so long.

At the first sight, it looks like the function ```timeout_to_seconds``` is a good place to add the support, but in fact it is not ideal. The issue is that this function is used for two things: 1) converting time strings to seconds 2) validation of time strings. In case of validation we need to just validate a value from the user and if it is valid then put it into the configuration. By allowing iso duration strings in this function we can run into a case where we accidentally put iso duration string to pacemaker configuration where it is not supported (example is stonith-watchdog-timeout property)

Therefore I suggest creating a new validator `class ValueTimeIntervalOrDuration` similar to [class ValueTimeInterval](https://github.com/ClusterLabs/pcs/blob/58a7306c7a7d78a737b83f4d8adc6df217d92d37/pcs/lib/validate.py#L993). It would validate iso duration string in case it starts with ‘P’ otherwise it would use `timeout_to_seconds` function. Message for allowed values could be: _"time interval in number+units or ISO8601 duration (e.g. 1, 2s, 3m, 4h, PT1H2M3S, ...)"_ Also new test coverage will be needed, take a look at [ValueTimeInterval test case](https://github.com/ClusterLabs/pcs/blob/58a7306c7a7d78a737b83f4d8adc6df217d92d37/pcs_test/tier0/lib/test_validate.py#L1726) to see how to do it.

New validator should be used in the [validate_set_cluster_properties](https://github.com/ClusterLabs/pcs/blob/58a7306c7a7d78a737b83f4d8adc6df217d92d37/pcs/lib/cluster_property.py#L166) function for all time properties except _stonith-watchdog-timeout_.

Regarding the concrete implementation:

The python library [isodate](https://pypi.org/project/isodate/) was chosen for the parsing of ISO durations. We discussed this decision and we came to the conclusion that we are not very comfortable with adding another python dependency to the project.

We came with an alternative solution which is to use an utility provided by pacemaker called [iso8601](https://clusterlabs.org/pacemaker/man/iso8601.8.html). Duration could be validated by calling command `iso8601 --duration=<duration_string>` and checking the exit code. An example of calling the pacemaker utility can be found in the function [push_cib_diff_xml](https://github.com/ClusterLabs/pcs/blob/58a7306c7a7d78a737b83f4d8adc6df217d92d37/pcs/lib/pacemaker/live.py#L246).  A function `def is_duration(runner: CommandRunner, value: str) -> bool` could be created in file [pcs/lib/pacemaker/values.py](https://github.com/ClusterLabs/pcs/blob/main/pcs/lib/pacemaker/values.py). This new function could be used in previously suggested validator `ValueTimeIntervalOrDuration`.

This suggested solution does not include conversion of iso duration values to number of seconds. It is not necessary for validation of time cluster properties. Also we don’t need to use these converted values at any palace in pcs. Especially converting years and months to seconds is challenging because they don’t always have the same number of days. Unfortunately neither of [isodate](https://pypi.org/project/isodate/) or [iso8601](https://clusterlabs.org/pacemaker/man/iso8601.8.html) provide convenient means to convert duration to seconds.

Please let us know if you have any questions or you have another opinion on this.

pcs/common/tools.py
mirecheck, line 1: Look at the comment in the comments section.

pcs/lib/cib/nvpair_multi.py
mirecheck, line 1: This place is not the right place for any validations. This layer should not have any knowledge about any specific nvpairs. All validation should be done before calling `nvset_update`. There is a function [validate_set_cluster_properties](https://github.com/ClusterLabs/pcs/blob/58a7306c7a7d78a737b83f4d8adc6df217d92d37/pcs/lib/cluster_property.py#L166) for validating cluster properties.

mirecheck (changes requested):

daaitudian:
Hi @mirecheck,
Thank you for making this suggestion. After a while, I will change this pr according to your suggestion.

PR #602, opened 2022-12-02
add type hints to pcs.lib.corosync

tomjelinek:


pcs/lib/corosync/node.py
```diff
@@ -39,46 +38,64 @@ def to_dto(self) -> CorosyncNodeAddressDto:
...
+    name: Optional[str]
+    addrs: Sequence[CorosyncNodeAddress]
+    nodeid: Optional[str]
 
-    def __new__(cls, name, addrs, nodeid):
-        return super().__new__(cls, name, tuple(addrs), nodeid)
+    def __post_init__(self) -> None:
+        # Make sure addrs is a tuple so it is not possible to change its items.
+        # Since the dataclass is frozen, object__setattr__ must be used.
+        object.__setattr__(self, "addrs", tuple(self.addrs))
```
ondrejmular, line 52: Is this really necessary? `Sequence` is immutable so mypy should make sure it's not being changed.

pcs/common/types.py
```diff
@@ -67,3 +67,10 @@ def from_str(cls, transport: str) -> "CorosyncTransportType":
             return cls(transport.upper())
         except ValueError:
             raise UnknownCorosyncTransportTypeException(transport) from None
+
+
+class CorosyncNodeAddressType(AutoNameEnum):
```
ondrejmular, line 72: In case we want to deserialise this, it should be added to dacite config in `pcs.common.interface.dto.from_dict`

pcs/lib/tools.py
```diff
@@ -97,7 +101,7 @@ def write_tmpfile(data, binary=False):
 def get_tmp_file(
     data: Union[None, bytes, str] = None,
     binary: bool = False,
-) -> Any:
+) -> Iterator[Union[IO[str], IO[bytes]]]:
```
ondrejmular, line 104: Based on https://github.com/python/typeshed/issues/2772 we should use `Generator` here

```suggestion
) -> Generator[Union[IO[str], IO[bytes]], None, None]:
```

pcs/lib/tools.py
```diff
@@ -112,24 +116,28 @@ def get_tmp_file(
 
 
 @contextmanager
-def get_tmp_cib(report_processor: reports.ReportProcessor, data: str) -> Any:
+def get_tmp_cib(
+    report_processor: reports.ReportProcessor, data: str
+) -> Iterator[IO[str]]:
```
ondrejmular, line 121:
```suggestion
) -> Generator[IO[str], None, None]:
```

pcs/lib/tools.py
```diff
@@ -76,13 +78,15 @@ def dict_to_environment_file(config_dict):
     return "".join(lines)
 
 
-def write_tmpfile(data, binary=False):
+def write_tmpfile(
+    data: Union[None, bytes, str], binary: bool = False
+) -> Union[IO[str], IO[bytes]]:
```
ondrejmular, line 83: By using [@typing.overload](https://docs.python.org/3/library/typing.html#typing.overload) we can avoid casting its result like in this [case](https://github.com/ClusterLabs/pcs/pull/602/files#diff-e0413fa653ac3d6f221059dbeaa40db262b65694eb74a3f5133cddc574ce467fR129)

ondrejmular (changes requested):

pcs/lib/corosync/config_facade.py
```diff
@@ -578,66 +609,82 @@ def set_totem_options(self, options):
...
         """
-        for quorum in self.config.get_sections("quorum"):
-            for device in quorum.get_sections("device"):
-                if device.get_attributes("model"):
-                    return True
-        return False
+        return bool(self._get_raw_quorum_device_model())
 
-    def get_quorum_device_model(self):
+    def get_quorum_device_model(self) -> str:
```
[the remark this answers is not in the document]
tomjelinek: The `has_quorum_device` method existed before I started adding type hints.

The proposed change may look like it simplifies things. But I don't think it's worth it. If `get_quorum_device_model` could return `None`, we would have to check for that everywhere it's used. Otherwise `mypy` would rightfully complain if we tried to use `get_quorum_device_model`'s return value anywhere where only `str` is allowed.

Having one method to check if quorum device is defined and another methods to get quorum device configuration looks better to me than having methods to get quorum device configuration or `None` in case no quorum device is defined and checking if they returned `None` or actual data everywhere they're used.

pcs/lib/corosync/node.py
tomjelinek: Right, `mypy` prevents modifying `Sequence` here. Let's rely on that.

pcs/common/corosync_conf.py
```diff
@@ -1,26 +1,29 @@
...
-    type: str  # TODO: create enum of addr types in pcs.lib.corosync.node
+    type: CorosyncNodeAddressType
 
 
 @dataclass(frozen=True)
 class CorosyncNodeDto(DataTransferObject):
     name: str
     nodeid: str
-    addrs: List[CorosyncNodeAddressDto]
+    addrs: Sequence[CorosyncNodeAddressDto]
```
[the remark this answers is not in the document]
tomjelinek: It works

pcs/lib/corosync/config_facade.py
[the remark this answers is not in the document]
ondrejmular: So I checked all `has_quorum_device` and `get_quorum_device_model` calls (in `main` branch as I assume there should not be significant differences with this branch). Every time `get_quorum_device_model` is called, `has_quorum_device` is called right before it with some error handling.

Therefore I don't think check for `None` would add any complexity as basically the same check is done everywhere already, only wrapped in a function call. Which means we need to call 2 separate functions for no real benefit.

Actually, I believe having `mypy` checking for not properly handling the case when there is no quorum device is much better then getting an `AssertionError` at runtime.

ondrejmular (approved):

pcs/lib/corosync/config_facade.py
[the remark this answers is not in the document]
tomjelinek: You are right, your proposed version is actually simpler.

PR #623, opened 2023-01-27
replace Iterable[str] type hints

tomjelinek:


pcs_test/tier0/test_lint.py
```diff
@@ -0,0 +1,40 @@
+import os
+import re
+from unittest import TestCase
+
+PACKAGE_DIR = os.path.realpath(
+    os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+)
```
ondrejmular, line 7: This will not work with `--installed` option. IMHO the easiest way to solve this would be to use `import pcs; pcs.__path__`

pcs_test/tier0/test_lint.py
```diff
@@ -0,0 +1,40 @@
...
+        ):
+            del dirnames
+            for fname in filenames:
+                if not fname.endswith(".py"):
+                    continue
+                file_path = os.path.join(dirpath, fname)
+                with open(file_path) as file:
+                    filedata = file.read()
+                    if re.search(
+                        r"\b(collection|iterable|sequence)\[[^][]*str",
```
ondrejmular, line 26: maybe this way it will be easier to read `\b(collection|iterable|sequence)\[[^\[\]]*str`

ondrejmular (changes requested):

tomjelinek:
retest this please

ondrejmular (approved):

PR #624, opened 2023-01-27
fix a race-condition when creating new auth token

ondrejmular:


pcs/common/file.py
```diff
@@ -92,6 +99,9 @@ def write(self, file_data: bytes, can_overwrite: bool = False) -> None:
         """
         raise NotImplementedError()
 
+    def update(self) -> ContextManager[BytesIO]:
```
tomjelinek, line 102: Perhaps add a doc text

pcs_test/tier0/common/test_file.py
```diff
@@ -379,6 +386,171 @@ def test_cannot_write(self, mock_chmod, mock_chown, mock_flock):
...
+        self.assert_success(
+            self.raw_file,
+            self._modify_no_truncate,
+            mock_chmod,
+            mock_chown,
+        )
+
+    def test_success_binary(self, mock_chown, mock_chmod):
+        self.assert_success(
+            self.raw_file,
```
tomjelinek, line 460:
```suggestion
            self.raw_file_binary,
```

pcs_test/tier0/common/test_file.py
```diff
@@ -379,6 +386,171 @@ def test_cannot_write(self, mock_chmod, mock_chown, mock_flock):
...
+            self._modify,
+            mock_chmod,
+            mock_chown,
+        )
+
+    def test_success_no_truncate_binary(self, mock_chown, mock_chmod):
+        self.assert_success(
+            self.raw_file,
```
tomjelinek, line 468:
```suggestion
            self.raw_file_binary,
```

tomjelinek (changes requested):

tomjelinek (approved):

PR #637, opened 2023-02-13
configure.ac: do not depend on daemons gem

lucaskanashiro:
daemons gem was apparently added only because it is a thin dependency. Now that thin was replaced with puma, there is no need to depend on daemons anymore.

Commit with the replace of thin with puma:

https://github.com/ClusterLabs/pcs/commit/aba40f267c0d7926c7e1cf62d3de817c359e4c3d


tomjelinek:
test this please

tomjelinek:
Thank you for the patch!

tomjelinek (approved):

tomjelinek:
test this please

PR #639, opened 2023-02-20
allow creating qdevice certs on the local node only

tomjelinek:


tomjelinek:
retest this please

pcs/lib/corosync/qdevice_net.py
```diff
@@ -108,9 +113,154 @@ def set_up_client_certificates(
...
+        return match.group(1) if match else None
+
+    if not client_initialized():
+        return False
+
+    # check if certificates are present
+    certutil_cmd = [
+        settings.certutil_executable,
+        "-d",
+        settings.corosync_qdevice_net_client_certs_dir,
```
ondrejmular, line 185: Option `-L` could be added here at it is added everywhere. And `certutil_cmd` can be `tuple` instead of `list`.

ondrejmular (changes requested):

ondrejmular (changes requested):
File `pcs_test/tier0/cli/test_quorum.py` is not specified in `EXTRA_DIST` in `pcs_test/Makefile.am` file.

ondrejmular (approved):

PR #648, opened 2023-03-01
 add meta option usage for pcs resource clone

F16shen:
fix https://bugzilla.redhat.com/show_bug.cgi?id=2168155
I'm not sure modify usage enough for this issue. The output looks like it adds meta properties for the clone, not the base resource. But to my understanding meta option is set for primitive resource

### DIFF
--- a/pcs/cli/resource/parse_args.py
+++ b/pcs/cli/resource/parse_args.py
@@ -6,6 +6,7 @@ from pcs.cli.common.parse_args import (
     group_by_keywords,
     prepare_options,
 )
+from pcs.cli.reports.output import deprecation_warning
 
 
 def parse_create_simple(arg_list):
--- a/pcs/cli/resource/parse_args.py
+++ b/pcs/cli/resource/parse_args.py
@@ -50,6 +50,12 @@ def parse_clone(arg_list, promotable=False):
         raise CmdLineInputError(
             "op settings must be changed on base resource, not the clone",
         )
+
+    if "meta" in groups:
+        deprecation_warning(
+            "option 'meta' is deprecated and will be removed in a future release."
+        )
+
     parts["meta"] = prepare_options(
         groups.get("options", []) + groups.get("meta", []),
     )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,10 +23,16 @@
     line
   - Specifying `--full` to show IDs of elements now shows IDs of nvpairs as well
 
+### Deprecated
+- Keyword `meta` in `pcs resource clone` and `pcs resource promotable` commands
+  (it wasn't doing anything anyway) ([rhbz#2168155], [ghpull#648])
+
 [ghissue#612]: https://github.com/ClusterLabs/pcs/issues/612
+[ghpull#648]: https://github.com/ClusterLabs/pcs/pull/648
 [rhbz#1423473]: https://bugzilla.redhat.com/show_bug.cgi?id=1423473
 [rhbz#1860626]: https://bugzilla.redhat.com/show_bug.cgi?id=1860626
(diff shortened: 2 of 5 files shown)


tomjelinek (changes requested):
The usage is correct, there is nothing to be fixed in there. What's wrong is parsing of `pcs resource clone` arguments. See [comment 1](https://bugzilla.redhat.com/show_bug.cgi?id=2168155#c1) in the original bugzilla.

tomjelinek:
test this please

pcs/cli/resource/parse_args.py
```diff
@@ -30,7 +30,7 @@ def parse_create_simple(arg_list):
 def parse_clone(arg_list, promotable=False):
     parts = {
         "clone_id": None,
-        "meta": {},
+        "clone_option": {},
```
tomjelinek, line 33: This change is unnecessary, the arguments parsed into this key are really meta attributes of a clone.

pcs/cli/resource/parse_args.py
```diff
@@ -50,16 +50,15 @@ def parse_clone(arg_list, promotable=False):
...
             "op settings must be changed on base resource, not the clone",
         )
-    parts["meta"] = prepare_options(
-        groups.get("options", []) + groups.get("meta", []),
-    )
+
+    if "meta" in groups:
+        raise CmdLineInputError(
+            "meta options is not supported in 'pcs resource clone' command"
+        )
```
tomjelinek, line 57: For now, we need to allow `meta` keyword to maintain backward compatibility. So we cannot raise an exception here. Instead, print a deprecation warning using `pcs.cli.reports.output.deprecation_warning`.

pcs/cli/resource/parse_args.py
```diff
@@ -50,16 +50,15 @@ def parse_clone(arg_list, promotable=False):
         raise CmdLineInputError(
             "op settings must be changed on base resource, not the clone",
         )
-    parts["meta"] = prepare_options(
-        groups.get("options", []) + groups.get("meta", []),
-    )
+
+    if "meta" in groups:
+        raise CmdLineInputError(
+            "meta options is not supported in 'pcs resource clone' command"
```
tomjelinek, line 56: This function doesn't know it is called from `pcs resource clone` command. In fact, it is being called from `pcs resource clone`, `pcs resource promotable` and `pcs resource update`. The message should not refer to any specific pcs command.

pcs/cli/resource/parse_args.py
```diff
@@ -50,16 +50,15 @@ def parse_clone(arg_list, promotable=False):
...
+        )
+
+    parts["clone_option"] = prepare_options(groups.get("options", []))
     if promotable:
-        if "promotable" in parts["meta"]:
-            raise CmdLineInputError(
-                "you cannot specify both promotable option and promotable "
-                "keyword"
-            )
```
tomjelinek, line 61: This check and the related exception is needed. It prevents command `pcs resource promotable resource-name promotable=false` to pass, as it is a mistake to create a promotable clone and set its promotable flag to false in one step.

tomjelinek (changes requested):
Thank you for updating the pull request. However, there are still issues which need to be resolved. Feel free to reach out to us if you need more guidance.

F16shen:
Thanks for the comments. 
I only add a deprecated warning this time. Should I add tests as well?

tomjelinek:
test this please

tomjelinek:
test this please

tomjelinek (approved):
Thanks, this was it.

I just made sure that the deprecation warning doesn't get printed from `pcs resource update` command. And I updated tests and changelog, as there was not much to do.

PR #662, opened 2023-03-16
[pcs-0.10] add check for closed stdin

CtrlZmaster:
Authors of scripts sometimes close stdin. In some commands, we ask the user for input but first we check that pcs is not invoked from a script. This check was returning a traceback when stdin was closed. This commit adds a check that stdin is not None - that indicates that stdin is closed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@
 - Validate dates in location constraint rules ([ghpull#644], [rhbz#2178707])
 - Fix displaying differences between configuration checkpoints in
   `pcs config checkpoint diff` command ([rhbz#2176490])
+- Crash in commands that ask for user input (like `pcs cluster destroy`) when
+  stdin is closed ([ghissue#612])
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
...
 ## [Unreleased]
 
+### Fixed
+- Crash in commands that ask for user input (like `pcs cluster destroy`) when
+  stdin is closed ([ghissue#612])
+
+[ghissue#612]: https://github.com/ClusterLabs/pcs/issues/612
+
+
+## [0.11.5] - 2023-03-01
```
ondrejmular, line 12: I guess this should not be here

ondrejmular (changes requested):

CHANGELOG.md
CtrlZmaster: Definitely not, not sure how I missed it. I'm going to fix it.

PR #676, opened 2023-04-21
fix 'pcs resource create' fails when the empty interval is specified for an operation

daaitudian:
Hi,
This pr is fixed bug [https://bugzilla.redhat.com/show_bug.cgi?id=2179084](url)


tomjelinek:
test this please

mirecheck (changes requested):
Hi @daaitudian,
thank you for your pull request and interest in our project!
Unfortunately it cannot by accepted as it is.  The main reasons are:
1. the place where it is fixed is not the right place to fix the issue
2. the fix add new report message instead of re-using existing report message

If you would like to improve the fix, here are guidelines:

1. Add validator [ValueTimeInterval](https://github.com/ClusterLabs/pcs/blob/0a2cb5ac9358e265000d3602ae575353dec9614d/pcs/lib/validate.py#L1024) for `interval` option into the function [_validate_operation_list()](https://github.com/ClusterLabs/pcs/blob/0a2cb5ac9358e265000d3602ae575353dec9614d/pcs/lib/cib/resource/operations.py#L172). Please add also the same validator for the `timeout` option, because its validation is missing too.

- this validator reports `INVALID_OPTION_VALUE` report message in case the value is empty or it is not a valid time interval value
2. extend test [test_validate_all_individual_options](https://github.com/ClusterLabs/pcs/blob/0a2cb5ac9358e265000d3602ae575353dec9614d/pcs_test/tier0/lib/cib/test_resource_operations.py#L306) with interval option which has empty value and timeout option with a wrong value.
3. Add tier1 test to [FailOrWarnOp](https://github.com/ClusterLabs/pcs/blob/0a2cb5ac9358e265000d3602ae575353dec9614d/pcs_test/tier0/lib/cib/test_resource_operations.py#L306) class with command which contains interval option without a value

I hope that explanations are clear. If you have any questions don't hesitate to ask in comments.

daaitudian:
Hi @mirecheck :
Thanks for your help. I have completed the modification. If you have any other suggestions, please tell me.

mirecheck (commented):
Hello @daaitudian,
your modificiation looks good. I have just added a changelog update and some minor fixes. The PR could be merged after CI is finished successfully.

mirecheck (approved):

tomjelinek (approved):

PR #712, opened 2023-07-10
Support ISO 8601 duration specifications for pacemaker "time" properties

daaitudian:
Hi @mirecheck,
I completed the adjustment based on your advice #598. 
I submitted the second commit [37c4dcc](https://github.com/ClusterLabs/pcs/pull/712/commits/37c4dcc689ceb91215454f7e77180477b53e9327) to prevent the ISO8601 format time is used for stonith-watchdog-timeout.

### DIFF
--- a/pcs/lib/pacemaker/values.py
+++ b/pcs/lib/pacemaker/values.py
@@ -23,6 +23,7 @@ _ID_FIRST_CHAR_NOT_RE = re.compile("[^a-zA-Z_]")
 _ID_REST_CHARS_NOT_RE = re.compile("[^a-zA-Z0-9_.-]")
 SCORE_INFINITY = "INFINITY"
 
+
 def __exec(name):
     return os.path.join(settings.pacemaker_binaries, name)
 
@@ -68,11 +69,7 @@ def is_score(value: str) -> bool:
 
 
 def is_duration(runner: CommandRunner, value: str) -> bool:
-    cmd = [
-        __exec("iso8601"),
-        "--duration",
-        value
-    ]
+    cmd = [__exec("iso8601"), "--duration", value]
     _, _, retval = runner.run(cmd)
     return retval == 0
 
--- a/pcs/lib/pacemaker/values.py
+++ b/pcs/lib/pacemaker/values.py
@@ -1,3 +1,4 @@
+import os
 import re
 from typing import (
     List,
@@ -5,6 +6,7 @@ from typing import (
     Union,
 )
 
+from pcs import settings
 from pcs.common import reports
 from pcs.common.reports.item import (
     ReportItem,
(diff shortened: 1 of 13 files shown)


pcs_test/tier0/lib/test_validate.py
```diff
@@ -1748,6 +1752,37 @@ def test_reports_about_invalid_interval(self):
...
+                fixture.error(
+                    reports.codes.INVALID_OPTION_VALUE,
+                    option_name="a",
+                    option_value="invalid_value",
+                    allowed_values="time interval (e.g. 1, 2s, 3m, 4h, PT1H2M3S, ...)",
+                    cannot_be_empty=False,
+                    forbidden_characters=None,
+                ),
+            ],
+        )
```
mirecheck, line 1783: Here is missing coverage for iso duration values starting with 'P'. It would be enough to re-work it as a subTest with two values ['invalid_value', 'P'] and calling [assert_called_once_with](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once_with) and [assert_not_called](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_not_called) on runner mock. Also runner mock needs to be set to return a non-zero value.

pcs/lib/pacemaker/values.py
```diff
@@ -61,6 +67,18 @@ def is_score(value: str) -> bool:
     return unsigned_value == SCORE_INFINITY or unsigned_value.isdigit()
 
 
+def is_duration(runner: CommandRunner, value: str) -> bool:
```
mirecheck, line 70: Please add test coverage for this function into file [pcs_test/tier0/lib/pacemaker/test_values.py](https://github.com/ClusterLabs/pcs/blob/main/pcs_test/tier0/lib/pacemaker/test_values.py). That would be two simple tests with mocked runner.

pcs/lib/validate.py
mirecheck, line 1: This second commit is not the way how to handle `stonith-watchdog-timeout` property. It should be done by keeping `ValidateTimeInterval` validator for `stonith-watchdog-timeout` property in [validate_set_cluster_properties](https://github.com/ClusterLabs/pcs/blob/9d0bdcd42d29b1d7ca783d8b7585401ba59fffa9/pcs/lib/cluster_property.py#L166C1-L175C14) function. This commit could be discarded.

mirecheck (changes requested):
Hi @daaitudian,
thank you for your effort you are putting into this pull request. We really appreciate it.
There are few changes that has to be made before applying this PR. Please check the review comments and let me know if something is not clear.

mirecheck:
retest this please

daaitudian:
Hi @mirecheck,
Thanks for your help, I completed the adjustment based on your advice.

mirecheck (commented):
Hi @daaitudian,
Thanks for the update! I have fixed tests and issues reported by pylint, mypy and black. The pull request could be merged once it passes CI.

mirecheck (approved):

PR #733, opened 2023-09-04
confusing error msg booth grant/revoke

F16shen:
fix https://bugzilla.redhat.com/show_bug.cgi?id=2232143

### DIFF
--- a/pcs/lib/commands/booth.py
+++ b/pcs/lib/commands/booth.py
@@ -630,10 +630,8 @@ def _ticket_operation(
 
     if not site_ip:
         site_error = LibraryError(
-                        ReportItem.error(
-                            reports.messages.BoothCannotDetermineLocalSiteIp()
-                        )
-                    )
+            ReportItem.error(reports.messages.BoothCannotDetermineLocalSiteIp())
+        )
 
         try:
             cib = env.get_cib()
@@ -641,8 +639,8 @@ def _ticket_operation(
             raise site_error from e
 
         site_ip_list = resource.find_bound_ip(
-                get_resources(cib), booth_env.config_path
-            )
+            get_resources(cib), booth_env.config_path
+        )
         if len(site_ip_list) != 1:
             raise site_error
 
--- a/pcs/lib/commands/booth.py
+++ b/pcs/lib/commands/booth.py
@@ -629,15 +629,23 @@ def _ticket_operation(
     _ensure_live_env(env, booth_env)
 
     if not site_ip:
+        site_error = LibraryError(
+                        ReportItem.error(
+                            reports.messages.BoothCannotDetermineLocalSiteIp()
+                        )
+                    )
+
+        try:
+            cib = env.get_cib()
+        except LibraryError as e:
+            raise site_error from e
+
         site_ip_list = resource.find_bound_ip(
-            get_resources(env.get_cib()), booth_env.config_path
-        )
-        if len(site_ip_list) != 1:
-            raise LibraryError(
-                ReportItem.error(
-                    reports.messages.BoothCannotDetermineLocalSiteIp()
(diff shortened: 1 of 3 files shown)


pcs/lib/commands/booth.py
```diff
@@ -629,6 +629,19 @@ def _ticket_operation(
...
+
+        booth_status = stdout.rstrip()
+        booth_type = booth_status.split(" ")[4].split("=")[1]
+        if booth_type == "arbitrator":
+            raise LibraryError(
+                ReportItem.error(
+                    reports.messages.BoothSiteIpNotSpecifiedOnArbitrator()
+                )
+            )
+
```
mirecheck, line 644: This solution is a little bit complex and report message is too specific for arbitrator. 
The main issue is that it raises exception `IndexError: list index out of range` in case there is empty output from `booth status` command which may happen if there is no running booth daemon on the arbitrator node or there is no booth resource running on cluster node.

I would suggest to wrap `env.get_cib()` call to `try ... except` block and in case of `LibraryError` raise another `LibraryErrror` with the existing error message `BoothCannotDetermineLocalSiteIp`. 

This change would also require update to the unit tests. Please look at [test_cannot_find_site_ip](https://github.com/ClusterLabs/pcs/blob/f8381657a52615145566454094f3c13a0a3e9666/pcs_test/tier0/lib/commands/test_booth.py#L2311) test and add a new one named `test_cannot_load_cib`.  It will be almost the same as `test_cannot_find_site_ip` but the difference will be in line with  `self.config.runner.cib.load()` call. A non-zero returncode parameter and some stderr parameter need to be added.

pcs/common/reports/messages.py
```diff
@@ -6858,6 +6858,19 @@ def message(self) -> str:
...
+    booth site must be specified when granting tickets from an arbitrator.
+    """
+
+    _code = codes.BOOTH_SITE_IP_NOT_SPECIFIED_ON_ARBITRATOR
+
+    @property
+    def message(self) -> str:
+        return "Unable to operate booth ticket on arbitator node without specifying site ip"
+
+
```
mirecheck, line 6873: No need for a new report message. There is existing report message `BoothCannotDetermineLocalSiteIp` which does not mention arbitrator node and can be used for other cases when there is no cib available on a node.

mirecheck (changes requested):
Hello @F16shen!
Thank you for your pull request we appreciate it. There is some issues with the pull request and some changes are needed in order to accept your pull request. Please, see the review comments.

Thank you once again for your time and effort.

pcs/lib/commands/booth.py
```diff
@@ -629,15 +629,23 @@ def _ticket_operation(
...
+        except LibraryError as e:
             raise LibraryError(
-                ReportItem.error(
-                    reports.messages.BoothCannotDetermineLocalSiteIp()
+                    ReportItem.error(
+                        reports.messages.BoothCannotDetermineLocalSiteIp()
+                    )
                 )
-            )
+
```
mirecheck, line 648: Please put this BoothCannotDetermineLocalSiteIp LibraryError into variable `site_error` because it is used twice and use `try .. except` block only for getting cib:
```
try:
    cib = env.get_cib()
except LibraryError as e:
    raise site_error from e
```

Please do the same for pcs-0.10.

mirecheck (changes requested):
Modifications looks better but small code improvements could be done. Please, look at the review comments.
Thank you for your work.

F16shen:
Hi @mirecheck !
Thanks for the detailed explanation, the code look cleaner now.

mirecheck (approved):
I have added changelong and fixed linter issues. PR could be merged after successful CI pass.

PR #793, opened 2024-03-27
Deprecate pcmk removal

tomjelinek:


pcs_test/tier0/common/reports/test_messages.py
```diff
@@ -369,22 +369,50 @@ def test_cannot_be_empty_and_forbidden_characters(self):
...
+    def test_to_be_removed_no_replacement(self):
+        self.assert_message_from_report(
+            "option 'option name' is deprecated and will be removed",
+            reports.DeprecatedOption("option name", [], "", True),
+        )
+
+    def test_desc_component_removed_no_replacement(self):
+        self.assert_message_from_report(
+            "option type option 'option name' is deprecated by pacemaker and "
+            "will be removed",
```
ondrejmular, line 397: Looking at this message, I'm not sure if users will know what is 'pacemaker' in this context. Maybe we should be more specific, like 'pacemaker package', 'pacemaker project' or something similar which will be more explanatory for the users.

ondrejmular (approved):

ondrejmular (approved):

PR #803, opened 2024-04-29
Remove ms

tomjelinek:


mirecheck (changes requested):
Please update function update_resource in pcsd/remote.rb. There are options group, after, before used in deprecated way with '--'.

Otherwise, I haven't found any issues.

mirecheck (approved):

tomjelinek:
> Please update function update_resource in pcsd/remote.rb. There are options group, after, before used in deprecated way with '--'.
> 
> Otherwise, I haven't found any issues.

This is planned to be resolved in the future. It is actually unrelated to this PR.

PR #828, opened 2024-06-04
validate sbd options

tomjelinek:


pcs/lib/commands/sbd.py
```diff
@@ -56,6 +60,15 @@ def __tuple(set1, set2):
 )
 
 
+class _ValueSbdDelayStart(validate.ValuePredicateBase):
+    def _is_valid(self, value: validate.TypeOptionValue) -> bool:
+        # 1 means yes, so we don't allow it to prevent confusion
+        return value in ["yes", "no"] or is_integer(value, 2)
```
mirecheck, line 66: wouldn't be better to use [is_true](https://github.com/ClusterLabs/pcs/blob/02a5304fb306d86cd3b8bf89b324fd34ab8e1f20/pcs/lib/pacemaker/values.py#L37) function from pcs.lib.pacemaker.values?
Sbd uses [crm_is_true](https://github.com/ClusterLabs/sbd/blob/42455eb4298fa4eae03330ca1a1bc373ece58c8f/src/sbd-inquisitor.c#L991C28-L991C29)
```suggestion
        return is_true(value) or is_integer(value, 2)
```

mirecheck (changes requested):

pcs/lib/commands/sbd.py
tomjelinek: SBD man page says that only "yes", "no" and an integer is allowed. The other values are undocumented, and so they can get removed anytime. We'll stick to officially documented and supported values.

mirecheck (approved):

PR #877, opened 2024-10-21
Fix remote and guest node detection in resource delete

Pederrr:


tomjelinek (changes requested):
The change looks good overall. However, the line `_ensure_not_guest_remote(elements_to_remove.resources_to_disable)` feels weird, even though it works correctly. It would be better to have `_ensure_not_guest_remote(elements_to_remove.resources_to_remove)`. Also, it would be nice if properties of `ElementsToRemove` were commented, so that it is clear what they contain. Thanks.

tomjelinek (approved):

PR #889, opened 2024-11-07
Allow custom gemfile

mirecheck:


tomjelinek (changes requested):
The behavior should be controlled by a switch passed to configure script. When it's controlled by the presence of a file, it is rather nontransparent and nonintuitive.

tomjelinek (approved):

PR #895, opened 2024-11-20
Allow rpmbuild in venv for CI [DO NOT MERGE]

CtrlZmaster:
This patch should allow upstream CI to run pcs-0.12 on RHEL 9 builders if configure is ran with `--with-python-version=3.12`.


fabbione:
Superseeded by #900

PR #1003, opened 2025-07-29
Do not require wheel when setuptools is new enough

hroncok:
The current version of setuptools creates wheels on its own.

[In Fedora ELN, I'd like to get rid of wheel](https://github.com/fedora-eln/eln/issues/284).

This is my untested attempt to get rid of the dependency. I don't do much m4 nowadays, so forgive me if it's not correct. I am opening this as a draft to see what the CI says.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Apart from the dependencies listed above, these are also required for
 installation:
 
 * python development files (packages python3-devel, python3-setuptools 66.1+,
-  python3-pip, python3-wheel)
+  python3-pip, python3-wheel if python3-setuptools < 71)
 * ruby development files (package ruby-devel)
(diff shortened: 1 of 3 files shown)


CtrlZmaster:
@hroncok I'm ok with merging this pending Fedora checks here: https://src.fedoraproject.org/rpms/pcs/pull-request/58# Is that ok with you?

Note aside: Ooops, I accidentally pushed to rawhide instead of my fork, so the MR in Fedora is actually already merged :cold_sweat:

CtrlZmaster:
The Fedora downstream checks succeeded too: https://src.fedoraproject.org/rpms/pcs/pull-request/58#comment-274708
Merging in 3...2...1...

PR #1107, opened 2026-04-20
Handle unknown enum values in APIv2

Pederrr:
Closes #1095 

Curl command to test fix:
```
$ curl -k -X POST http://192.168.122.3:2224/remote/set_permissions \
    --unix-socket /var/run/pcsd.socket \
    --data-urlencode 'json_data={
      "permissions": {
        "0": {"name": "foo", "type": "user", "allow": {"BAD": "1"}}
      }
    }'
```

The fix also updates existing test to fail when APIv2 command uses Enum in parameters, and there is no dacite typehook for the Enum.

### DIFF
--- a/pcs/common/interface/dto.py
+++ b/pcs/common/interface/dto.py
@@ -36,12 +36,22 @@ SerializableType = Union[
 ]
 
 T = TypeVar("T")
+E = TypeVar("E", bound=Enum)
 
 ToDictMetaKey = NewType("ToDictMetaKey", str)
 META_NAME = ToDictMetaKey("META_NAME")
 
 
-E = TypeVar("E", bound=Enum)
+class PayloadConversionError(Exception):
+    pass
+
+
+class _UnionNotAllowed(Exception):
+    pass
+
+
+class DataTransferObject(DataclassInstance):
+    pass
 
 
 def _safe_enum_cast(enum_class: type[E]) -> Callable[[Any], E]:
@@ -52,7 +62,7 @@ def _safe_enum_cast(enum_class: type[E]) -> Callable[[Any], E]:
             valid_values = format_list([f.value for f in enum_class])
             raise PayloadConversionError(
                 f"Invalid value '{value}' for Enum '{enum_class.__name__}', "
-                f"use one of {valid_values}"
+                f"expected one of {valid_values}"
             ) from e
 
     return _cast_value
@@ -97,18 +107,6 @@ DTO_TYPE_HOOKS_MAP: dict[type[Any], Callable[[Any], Any]] = {
     #   queries: Sequence[tuple[str, str]]
     tuple[str, str]: lambda v: tuple(v) if isinstance(v, list) else v,
 }
-
-
-class PayloadConversionError(Exception):
-    pass
-
-
-class _UnionNotAllowed(Exception):
-    pass
-
-
-class DataTransferObject(DataclassInstance):
-    pass
 
 
 def meta(name: str) -> dict[str, str]:
--- a/pcs/common/interface/dto.py
+++ b/pcs/common/interface/dto.py
@@ -19,6 +19,7 @@ import dacite
 import pcs.common.async_tasks.types as async_tasks_types
 import pcs.common.permissions.types as permissions_types
 from pcs.common import types
+from pcs.common.str_tools import format_list
 
 if TYPE_CHECKING:
     from _typeshed import DataclassInstance  # pylint: disable=import-error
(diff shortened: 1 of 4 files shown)


pcs_test/tier0/daemon/async_tasks/test_command_mapping.py
```diff
@@ -52,27 +55,39 @@ def _find_disallowed_types(_type, allowed_types, hooked_origins, _seen=None):
...
+    else:
+        # flag as disallowed if types' origin needs a hook but this specific
+        # instantiation does not have a type hook
+        # e.g. we found a tuple, that does not have explicit type hook, so
+        # we flag it as disallowed
+        if (
+            generic in hooked_generic_origins
+            and Ellipsis not in get_args(_type)
+            and _type not in allowed_types
+        ):
```
Pederrr, line 71: The test was originally written to extract the generic origins from the types in type hooks (originally, only `tuple`), and then fail if there is any argument with that type origin, that does not have explicit typehook. This works, because there are only typehooks with `tuple` generic origin, and tuples must have explicit type hooks in our case. I kept the functionality the same.

However, if we in the future for whatever reason needed a new typehook for e.g. `list[MySpecialType]`, then the test will start failing for all lists, because it extracts the `list` generic type origin, and this `if` will be `True` for any argument that is `list`.

Is this approach really a good idea? Or would it be better to just say that tuple is not allowed in this test?

pcs_test/tier0/daemon/async_tasks/test_command_mapping.py
```diff
@@ -98,15 +113,19 @@ def test_all(self):
 
     def test_check_type_hooks_map_types_in_commands(self):
         allowed_types = set(DTO_TYPE_HOOKS_MAP.keys())
-        hooked_origins = {get_origin(_type) for _type in allowed_types}
+        hooked_generic_origins = {
+            origin
+            for _type in allowed_types
+            if (origin := get_origin(_type)) is not None
+        }
```
tomjelinek, line 120: This comprehension is based on the fact that `get_origin` returns `None` for types not supporting arguments, effectively filtering out primitive types. This may not be obvious, as it is not a primary goal of `get_origin`. Thus it deserves an explanatory comment stating what `get_origin` does and how it is used here.

pcs/common/interface/dto.py
```diff
@@ -40,7 +41,47 @@ class DataclassInstance:
...
+E = TypeVar("E", bound=Enum)
+
+
+def _safe_enum_cast(enum_class: type[E]) -> Callable[[Any], E]:
+    def _cast_value(value: Any) -> E:
+        try:
+            return enum_class(value)
+        except ValueError as e:
+            valid_values = format_list([f.value for f in enum_class])
+            raise PayloadConversionError(
```
tomjelinek, line 53: `PayloadConversionError` is referenced before it's defined. This works at runtime because the reference is inside a closure that's only called later. However, I believe the file would be a bit easier to read if `PayloadConversionError`, `_UnionNotAllowed` and `DataTransferObject` definitions were moved above `_safe_enum_cast`.

pcs/common/interface/dto.py
```diff
@@ -40,7 +41,47 @@ class DataclassInstance:
...
+                f"Invalid value '{value}' for Enum '{enum_class.__name__}', "
+                f"use one of {valid_values}"
```
tomjelinek, line 55:
```suggestion
                f"expected one of {valid_values}"
```
Just a tiny wording improvement.

pcs/daemon/async_tasks/worker/executor.py
tomjelinek, line 215:
```suggestion
        logger.error("Task %s raised a LibraryError: %s.", task.task_ident, e)
```
Using `logger.exception` puts the whole traceback associated with the exception into daemon log, which makes the log grow fast on invalid inputs. Since `LibraryError` contains reports clearly pointing to the origin of the exception, the traceback brings very little additional information and rather pollutes the log. Let's use `logger.error` instead and log the report messages if any.

pcs/daemon/async_tasks/worker/executor.py
```diff
@@ -195,7 +195,7 @@ def task_executor(task: WorkerCommand) -> None:
         except dto.PayloadConversionError as e:
             raise LibraryError(
                 reports.ReportItem.error(
-                    reports.messages.CommandInvalidPayload("")
+                    reports.messages.CommandInvalidPayload(str(e))
                 )
             ) from e
```
tomjelinek, line 200: This `except` block is the same as the previous one, so they can be merged.

pcs_test/tier0/common/interface/test_dto.py
```diff
@@ -68,6 +69,14 @@ class TypeHooksDto(DataTransferObject):
     optional_tuple_str_str: Optional[tuple[str, str]]
 
 
+@dataclass
+class EnumDto(DataTransferObject):
```
tomjelinek, line 73: This class is only useful for `FromDictEnumConversion` test case. Move it right above it so that related code is placed close to each other to make it easier to read.

pcs_test/tier0/daemon/async_tasks/test_command_mapping.py
tomjelinek: Good catch. The condition (and the whole test) looks like it is written generically, but it checks for Ellipsis, which is used specifically in `tuple`s. How do you propose to handle this?

tomjelinek (changes requested):
Good job. I found a few issues, but they are mostly of low impact.

pcs_test/tier0/daemon/async_tasks/test_command_mapping.py
Pederrr: I think the test can be simplified to check specifically for tuples only in this case, and not to try to work generically depending on what generic types are in the DTO_TYPE_HOOKS_MAP. 

As I said, it works now because there are only tuples in the type hooks, but it would start failing for e.g. all lists if we added a type hook for just one specific type of list. The upside of the current approach is that if we find out we need type hook for e.g. every `set`, then the test will 🪄 automatically 🪄 work after we add a type hook for any `set` (and properly fail for any other `set` without type hook). But the test is kind of complicated, and I think it will be easier to add another `if isintance(...)` check when we realize that we actually need a type hook, than trying to fix the test when someone adds a typehook that shouldn't be generalized.

So, I am voting to simplify the test and make it check for only tuples (and enums of course). What do you think about this?

tomjelinek: Agreed, that's my thinking as well. Please proceed with simplifying the test.

pcs/daemon/async_tasks/worker/executor.py
Pederrr: I don't exactly get what do you mean by "log the report messages if any" - do you want to log all of the reports that were created in the command - if so, I am starting to think this is kind of out of scope of the original fix.

pcs_test/tier0/daemon/async_tasks/test_command_mapping.py
Pederrr: This code got removed as we simplified the test.

pcs/daemon/async_tasks/worker/executor.py
tomjelinek: > do you want to log all of the reports that were created in the command

No. Only log report messages included in the exception. My suggestion changes the behavior so that it logs the exception without the traceback. For example:

`ERROR -- : Worker#8319: Task ea1ad179-53e0-4061-9518-c27f46d6c95d raised a LibraryError: ReportItem(severity=ReportItemSeverity(level='ERROR', force_code=None), message=CommandInvalidPayload(reason="Invalid value 'BAD' for Enum 'PermissionGrantedType', use one of 'full', 'grant', 'read', 'write'"), context=None).`

This is good enough, the only change is the missing traceback, which is the main point. Optionally, we could properly render report items (`report_item.message.message`).

Pederrr: All right, I will look at this

Pederrr: Oh, I am sorry. I misunderstood, and didn't notice we actually log the report in your suggestion.

tomjelinek (approved):

tomjelinek (approved):

PR #1125, opened 2026-05-14
Apiv2 fix unions

Pederrr:
Closes: #1096 

We analyzed the code that caused the issues, and realized multiple things. 

The original purpose of the code in `pcs/common/interface/dto.py` was to rename dataclass attributes according to their META_NAME meta attribute. Later, enum translation was added in commit https://github.com/ClusterLabs/pcs/commit/9e6134d073fb6a2c09cdd19e2122f65449354480, which tried to fix issues with serialization of enum values. However, we analyzed the code and realized:

- META_NAME renaming is used only in one DTO, and this renaming mechanism is not even used for serialization and deserialization of that DTO
- This mechanism does not work correctly in APIv2: It translates the data based on type hints, but the return value of library commands is typed as `any` in DTO for APIv2 result (`pcs/common/async_tasks/dto.py`), which means that any transformations are skipped

This means that the code was a complicated piece of dead code and we decided that the best course of action is to remove it, and simplify the serialization and deserialization functions. This fixes the issue described in #1096 

We also realized, that we do not need to transform Enums into their values during serialization, as json can transform Enums that are also inheriting from a primitive type. There were however changes in Python 3.11 (https://blog.pecar.me/python-enum/) that caused the original issue that the https://github.com/ClusterLabs/pcs/commit/9e6134d073fb6a2c09cdd19e2122f65449354480 tried to solve. The simpler solution is to replace all instances of inheriting from `str, enum.Enum` to inherit from `enum.StrEnum`.

### DIFF
--- a/pcs/common/interface/dto.py
+++ b/pcs/common/interface/dto.py
@@ -105,25 +105,8 @@ def from_dict(
     )
 
 
-# TODO:
-# This actually might not be needed, if we make sure all Enums in the code
-# are (str, Enum) or (int, Enum) or similar. In that case, json.dumps
-# will transform the values correctly.
-def _enum_converter_factory(data: list[tuple[str, Any]]) -> dict[str, Any]:
-    def convert(obj: Any) -> Any:
-        if isinstance(obj, Enum):
-            return obj.value
-        if isinstance(obj, (list, tuple)):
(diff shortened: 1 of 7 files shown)


tomjelinek (approved):
LGTM

PR #1144, opened 2026-06-24
Replace deprecated aliases from typing module

nrwahl2:
pcs is using a bunch of aliases from the typing module that were deprecated in Python 3.9. This PR gets rid of those uses. It also replaces uses of `typing.Union` and `typing.Optional` with `|` syntax, which became available in Python 3.10.

Please feel free to take this over and file your own PR using these commits with your modifications, if it's easier than engaging in a review cycle with me.


tomjelinek:
retest this please

tomjelinek:
@nrwahl2 Thank you for the PR!

